PR #1, opened 2015-02-23
Prefer ActiveRecord methods to raw SQL

stevegrossi:
Check it out: a pull request against our styleguide! We can keep track of the discussion, and if it gets merged, the hosted styleguide will automatically reflect the change. Yay Jekyll!


armilam:
I think these rules make a lot of sense. I also like the wording that states that it's okay to use raw SQL, but you should use ActiveRecord querying if possible

armilam:
Just noticed something in the capitalize SQL section below: in the "so good!" section, the `on` is not capitalize it. Could you capitalize it while you're here, please?

PR #2, opened 2015-03-17
Extract a separate page for Rails-related style rules

stevegrossi:
Just a cut-n-paste job, plus making a few apostrophes use correct punctuation (`’` over `'`).


armilam:
:shipit:

PR #3, opened 2015-03-19
Reference bbatsov’s Rails styleguide

stevegrossi:
Note that it’s a good reference, though we don’t necessarily follow it explicitly.


armilam:
:shipit: !

PR #4, opened 2015-03-30
Use ActiveRecord  methods only for booleans

armilam:
Yep!

### DIFF
--- a/rails-styleguide.md
+++ b/rails-styleguide.md
@@ -46,4 +46,4 @@ While we don’t follow it explicitly, the [community Rails Styleguide](https://
       progress.score.present?   # returns true
 
       # booleans are okay
-      company.assignment_notifications_disabled?  #returns the actual value
+      company.assignment_notifications_disabled?  # returns the actual value
--- a/rails-styleguide.md
+++ b/rails-styleguide.md
@@ -34,13 +34,16 @@ While we don’t follow it explicitly, the [community Rails Styleguide](https://
 
 ### ActiveRecord shortcuts
 
-- Use `?` methods only for boolean values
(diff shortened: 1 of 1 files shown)


rails-styleguide.md
```diff
@@ -31,3 +31,16 @@ While we don’t follow it explicitly, the [community Rails Styleguide](https://
 
       # so good!
       Model.joins("LEFT JOIN things ON thing_id = things.id AND another_id = things.another_id")
+
+### ActiveRecord shortcuts
+
+- Use `?` methods only for boolean values
```
stevegrossi: When possible, I think we should give reasons for rules like this. How about

> ...to avoid unexpected behavior, such as ActiveRecord treating `0` as `false`, while Ruby considers it `true`.

stevegrossi:
Small suggestion :point_up:.

armilam:
How's it look now?

rails-styleguide.md
```diff
@@ -31,3 +31,19 @@ While we don’t follow it explicitly, the [community Rails Styleguide](https://
...
+      progress = Progress.create! lesson: a_lesson, user: a_user, score: 0.0
+
+      # do not do this to determine if the object has a value for score
+      progress.score?           # returns false
+
+      # do this
+      progress.score.present?   # returns true
+
+      # booleans are okay
+      company.assignment_notifications_disabled?  #returns the actual value
```
stevegrossi: Love the extra clarity. Mind adding a space between the `#` and the comment text here and it'll be good to ship?

armilam: Done!

stevegrossi:
Great! Thanks for adding this...

PR #6, opened 2015-04-13
Start a JS styleguide

stevegrossi:
Includes two things which came up recently in code reviews:

- use lowerCamelCase
- attach behavior to DOM elements with `data-behavior` attributes


armilam:
Awesome!

PR #7, opened 2015-04-15
Specify using lower-hyphen-case in Sass code

stevegrossi:
Since CSS already uses this case, we should be consistent (we currently use lowerCamelCase, snake_case, *and* lower-hyphen-case for mixins!)

@armilam, @kpascascio: can I get an amen?


armilam:
Amen! @kpascascio?

kpascascio:
amen

armilam:
Great! Merging!

PR #11, opened 2015-04-23
Add RSpec syntax (especially spacing) to testing styleguide

stevegrossi:
I'm not saying these are the best spacing rules, but we should be consistent. If you disagree about any of these, let's discuss!


stevegrossi:
@armilam, @kpascascio, thoughts?

armilam:
This looks good to me.

PR #12, opened 2015-05-26
Ruby style: use full words instead of letters for clarity

stevegrossi:
The particular bit of code that prompted this came from BulkAssignment.rb. Something like:

```ruby
@assignee.users.each do |user|
  assignables.map do |a|
    assignment = a.assign_to(user)
```

Here, we’re dealing with an assignee, assignables, and assignments, so it’s particularly unclear what `|a|` is referring to. But in general, I think this is a good practice. I’m surprised neither Bbatsov’s nor Github’s styleguides mention this, as I swear I’ve read it elsewhere.

### DIFF
--- a/ruby-styleguide.md
+++ b/ruby-styleguide.md
@@ -15,6 +15,13 @@ These override either Github’s or Batsov’s styleguide where applicable:
 - Use the Ruby 1.9-style hash syntax (`key: 'value'`) with symbol keys wherever possible. Only use `"hash" => "rocket"` syntax when hash keys are required to be strings.
 - Use a single newline above and below "private" and "protected" in classes
 - Use a single space within ERB tags, e.g. `<%= "foo" %>`, not `<%="foo"%>`
+- Use full words for variables except where a single letter is conventional (e.g. `f` for Rails form objects, `i` for iterators, etc.).
+      
+      # not so clear
+      assignee.assignments.map(&:assignable).each { |a| ...
+      
+      # so clear!
+      assignee.assignments.map(&:assignable).each { |assignable| ...
 
 ## Gemfile
 


stevegrossi:
@kpascascio and @kyprifog, your thoughts as well?

kpascascio:
I'm fine with it, if we come across a block just using letters should we change along with our PR's?

stevegrossi:
If this PR is accepted, then yeah @kpascascio, I'd say update things as you come across them, but no need to go out of our way. Especially with untested code, it's probably best not to change anything until we start working on/adding tests for it.

kpascascio:
@stevegrossi yup that's understandable!

ruby-styleguide.md
```diff
@@ -15,6 +15,13 @@ These override either Github’s or Batsov’s styleguide where applicable:
 - Use the Ruby 1.9-style hash syntax (`key: 'value'`) with symbol keys wherever possible. Only use `"hash" => "rocket"` syntax when hash keys are required to be strings.
 - Use a single newline above and below "private" and "protected" in classes
 - Use a single space within ERB tags, e.g. `<%= "foo" %>`, not `<%="foo"%>`
+- Use full words for variables except where a single letter is conventional (e.g. `f` for Rails form objects, `i` for iterators, etc.).
+      
+      # not so clear
+      assignee.assignments.map(&:assignable).each { |a| ...
```
armilam: These two lines are combined into one in the preview (same with the lines below).  Not sure if this is an artifact of previewing on a commit. Does this need a fix?

armilam:
I think this is a great idea. I did have one comment about the formatting, though it may not actually be a problem. Shoot it back at me if so.

ruby-styleguide.md
stevegrossi: Unfortunately, Jekyll and Github use different markdown renderers, so that seems to be an artifact of the commit preview. Here's how it looks with `$ jekyll serve`:

screen shot 2015-05-28 at 7 46 15 am

I'd like to change the styleguide to use Github-flavored markdown (especially for its fenced code blocks with triple-backticks), so hopefully this won't be an issue in the future.

PR #13, opened 2015-06-09
Avoid mixing inline and multiline conditionals

stevegrossi:
I noticed this one twice today. Since Github's preview doesn't parse our markdown right, here's how it looks:

screen shot 2015-06-09 at 4 38 41 pm


armilam:
I agree!

PR #14, opened 2015-06-15
Add Component-based CSS to CSS Stylegude

stevegrossi:
This PR attempts to codify our use of component-based CSS (so that all the new CSS added by Studio Science will be consistent not with the hodgepodge we have, but the architecture we *want*). We already use CSS components in many places (just look in `app/assets/stylesheets/components`), but much of our CSS remains unintentional. In other words, with selectors as long and as specific as necessary, the length and specificity of which tends to escalate over time, leading to larger CSS file sizes, poorer rendering performance, and difficulty maintaining it. (For more on the downsides of unscalable CSS, [see my blog post](http://work.stevegrossi.com/2014/09/06/how-to-write-css-that-scales/).) This is how we get selectors like:

```css
article.lesson_editor > aside ul.lesson_sections > li.active .lesson_section_title
```

I'd like to make the case for doubling down on component-based CSS which would let this instead be simply

```css
.StackedNav-item--active
```

### SuitCSS

What this PR does introduce that's new is the set of [SuitCSS naming conventions](https://suitcss.github.io/) for components. This adds—at no cost—an extra layer of semantics to our component names. Currently we generally use hyphen-separated lowercase class names, e.g. `.progress-bar`. In this case, it's impossible to tell whether the component is `.progress` and this is its bar element, or this is the base `.progress-bar` component. With SuitCSS, this would be named `.Progress-bar` and we'd immediately know just by looking at the markup that this is the bar element within the `.Progress` component. We have to name things anyway: we might as well name them in such a way as to provide the greatest amount of context and direction to future developers. Twitter and [Medium](http://work.stevegrossi.com/2014/09/06/how-to-write-css-that-scales/) have begun using SuitCSS, and most apps these days are using some kind of component-based naming convention, if not SuitCSS then [BEM](https://en.bem.info/) or [SmaCSS](https://smacss.com/book/categorizing).

### DIFF
--- a/css-styleguide.md
+++ b/css-styleguide.md
@@ -4,7 +4,60 @@ title: CSS Style Guide
 permalink: /css/
 ---
 
-In general, follow [Github's CSS styleguide](https://github.com/styleguide/css).
+## Structure CSS Around Components
+
+Also called "object-oriented CSS", component-based CSS brings a number of benefits to customer and developer happiness (in the form of improved performance and maintainability). Fundamentally, it's about encapsulating styles into self-contained units. For example, here is some traditional CSS:
+
+    // app/stylesheets/lessons.scss
+    #lesson_viewer {
+      background: #eee;
+      h1 {
+        font-size: 2em;
+      }
+      &.past_due {
+        background: #f00;
+      }
+      &.expanded {
+        height: 400px;
+      }
+    }
+
+and here is how it would be written in a component-based way (using [SuitCSS naming conventions](https://github.com/suitcss/suit/blob/master/doc/naming-conventions.md)):
+
+    // app/stylesheets/components/lesson_viewer.scss
+    .LessonViewer {
+      background: #eee;
+    }
+    .LessonViewer-heading {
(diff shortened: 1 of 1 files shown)


armilam:
I like this addition to the styleguide because:

- It reduces selector nesting a lot, which reduces load times and confusion from specificity
- Its naming convention makes it more clear as to what exactly the classes are for
- It encourages CSS to be defined not on a page-by-page basis, but rather by individual components and their own structures
- And more, but I think those reasons are enough

@TheAustinSeven, @kpascascio, and @kyprifog, check this PR out. I'd like your feedback as well.

kpascascio:
I like the shameless plus for website blog views :smile: @stevegrossi :+1: But I agree with this whole heartedly. All the projects that I've worked with CSS on I see how confusing and 'hidden' somethings are. I do want to read more into SuitCSS though!

armilam:
Thanks, @kpascascio! I'll go ahead and merge since we officially have majority vote :ballot_box_with_check:

PR #15, opened 2015-06-17
Document our usage of service objects with patterns

stevegrossi:
We're using this pattern a lot, to great effect. To familiarize new developers with it and to improve consistency in usage, I think it's worth adding to the styleguide.

Thoughts? @armilam, @kpascascio, @TheAustinSeven, @kyprifog

### DIFF
--- a/ruby-styleguide.md
+++ b/ruby-styleguide.md
@@ -38,7 +38,7 @@ These override either Github’s or Batsov’s styleguide where applicable:
 
 We use the service object pattern to encapsulate complex logic within self-contained objects. Instead of a 20-line controller action, we’ll call `DoSomethingComplicated.perform(args)` in the controller and move all of the complex logic into the `DoSomethingComplicated` class. This not only helps organize and isolate complex code, but also makes it easier to test.
 
-Service objects are called with `ServiceName.perform()`. They should be named imperatively after the actions they perform (e.g. `AssignUser`, `CreateGroup`).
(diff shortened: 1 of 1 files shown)


stevegrossi:
In [a comment on another PR](https://github.com/lessonly/lesson.ly/pull/1065/files#r32652975), @kyprifog brought up some valid points about DRYness here, specifically around how the `self.perform` method does basically the same thing across all services that instantiate objects. It'd be great to extract this into a base `Service` class that our services could inherit from and get the `self.perform` method for free.

armilam:
I think this is great. If we get agreement from the other devs, I'm happy to merge.

stevegrossi:
Yeah, no rush on these styleguide additions. Unlike app PRs, for these I think it's important that everyone have the time to provide input.

armilam:
@TheAustinSeven, @kpascascio, I'd like to point you guys as this again in case you forgot.

stevegrossi:
I'm going to make one small change on line 41, removing the parens from the method call:

```
# Service objects are called with `ServiceName.perform()`
Service objects are called with `ServiceName.perform`
```

I'd intended that to suggest that you can pass arguments to `ServiceName.perform`, but as written it's incompatible with our Ruby styleguide and potentially confusing.

TheAustinSeven:
I think it is very good, I agree that this should be merged.

PR #16, opened 2015-06-24
Explicitly use booleans in conditionals

armilam:
Just like the text says, this adds clarity to the purpose of the condition.


armilam:
@TheAustinSeven, @stevegrossi, a style guide addition for your consideration.

stevegrossi:
A fine practice! :+1:

TheAustinSeven:
I like this! It makes the code much easier to understand.

kpascascio:
Nooice!

PR #17, opened 2015-07-08
Prefer ActiveRecord bang methods

stevegrossi:
As discussed in our retrospective meeting, this will ensure we find out as early as possible when things fail for unexpected reasons.


stevegrossi:
cc: @lessonly/developers

armilam:
Yes!

kpascascio:
thanks for adding this snippet! It's good thing to have documented, heck it's just good to document things

PR #18, opened 2015-07-10
Prefer Time.current and Date.current

stevegrossi:
As recommended by the companion [Rails styleguide](https://github.com/bbatsov/rails-style-guide#time) to our Ruby styleguide. We don't explicitly follow the Rails styleguide, but we've been bitten by edge cases around this enough times that it's worth explicitly following.


stevegrossi:
cc: @lessonly/developers

kpascascio:
I love this idea, I thought we talked about this in the past when I was doing the timezone issue. But I'm glad to see this addition! 😊

armilam:
This is much needed. Thank you!

PR #19, opened 2015-07-27
Add stubbing-related guidelines

stevegrossi:
Check out the changes made for a description (they should in principle be clear enough—if not, I should probably reword.)

### DIFF
--- a/testing-styleguide.md
+++ b/testing-styleguide.md
@@ -25,6 +25,11 @@ Write tests for your code, please.
 
       expect(progress).to receive(:restart!).once
 
+## Guidelines
+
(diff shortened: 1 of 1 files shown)


stevegrossi:
Thoughts, @lessonly/developers?

armilam:
I agree with this word-for-word. I'll wait for feedback from @kpascascio and @TheAustinSeven before merging.

kpascascio:
I also agree with this, I got to see the change when I reviewed @stevegrossi PR that had this as an example.

armilam:
Well, we got one positive feedback and zero negative, so it's time to merge!

PR #21, opened 2015-09-02
Add a recommendation to freeze strings which should not change

stevegrossi:
We sometimes do this, but it looks like a best-practice. Here's a great article from the Honeybadger team on when and why to `freeze`: http://blog.honeybadger.io/when-to-use-freeze-and-frozen-in-ruby


armilam:
Great stuff!

PR #22, opened 2015-09-04
Prefer single quotes in Java/CoffeeScript

stevegrossi:
As noted in lessonly/lesson.ly#1433


armilam:
I like what you've done here.

PR #24, opened 2015-09-29
Don't reference instance variables in partials: prefer local variables

stevegrossi:
I'm running into this in the new overview pages: lessons and courses can share a lot of partials, but not when one expects `@lesson` and another expects `@course`. If these partials expected their dependencies to be passed in instead of magically being present, this would be simpler.

I think it's fine for a view template (e.g. `show.html.erb`) to reference instance vars (and be thus coupled) to its controller action, but it seems dangerous to couple *its own* dependents to the controller action as well.

```ruby
# Instead of
render "toolbar" # lessons/toolbar expects @lesson
render "toolbar" # courses/toolbar expects @course

# Use
render "shared/toolbar", resource: @lesson
render "shared/toolbar", resource: @course
```

### DIFF
--- a/rails-styleguide.md
+++ b/rails-styleguide.md
@@ -34,7 +34,7 @@ While we don’t follow it explicitly, the [community Rails Styleguide](https://
 
 ### Views
 
-- Don't reference instance variables in partials. It's alright for `show.html` to reference the `@lesson` declared in `LessonsController#show`, but if it calls out to `render "stats"`, `stats.html` should not know about `@lesson`. Instead pass it in directly: `render "stats", lesson: @lesson`. Eventually, we'll want to use that partial in another context where `@lesson` isn't defined: being explicit now saves us time later.
+- Avoid referencing instance variables in partials. It's alright for `show.html.erb` to reference the `@lesson` declared in `LessonsController#show`, but if it calls out to `render "stats"`, `_stats.html.erb` should not know about `@lesson`. Instead pass it in directly: `render "stats", lesson: @lesson`. Eventually, we'll want to use that partial in another context where `@lesson` isn't defined: being explicit now saves us time later.
 
 ### ActiveRecord shortcuts
 
--- a/rails-styleguide.md
+++ b/rails-styleguide.md
@@ -32,6 +32,10 @@ While we don’t follow it explicitly, the [community Rails Styleguide](https://
       # so good!
       Model.joins("LEFT JOIN things ON thing_id = things.id AND another_id = things.another_id")
 
+### Views
+
+- Don't reference instance variables in partials. It's alright for `show.html` to reference the `@lesson` declared in `LessonsController#show`, but if it calls out to `render "stats"`, `stats.html` should not know about `@lesson`. Instead pass it in directly: `render "stats", lesson: @lesson`. Eventually, we'll want to use that partial in another context where `@lesson` isn't defined: being explicit now saves us time later.
+
 ### ActiveRecord shortcuts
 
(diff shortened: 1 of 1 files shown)


stevegrossi:
@lessonly/developers, thoughts?

chuckshulicker:
Can you provide some context?

chuckshulicker:
https://github.com/lessonly/lesson.ly/commit/3c13b73aa5467df0b3d0b2cb9f3d861dfe473c3d :+1:

rails-styleguide.md
```diff
@@ -32,6 +32,10 @@ While we don’t follow it explicitly, the [community Rails Styleguide](https://
       # so good!
       Model.joins("LEFT JOIN things ON thing_id = things.id AND another_id = things.another_id")
 
+### Views
+
+- Don't reference instance variables in partials. It's alright for `show.html` to reference the `@lesson` declared in `LessonsController#show`, but if it calls out to `render "stats"`, `stats.html` should not know about `@lesson`. Instead pass it in directly: `render "stats", lesson: @lesson`. Eventually, we'll want to use that partial in another context where `@lesson` isn't defined: being explicit now saves us time later.
```
armilam: To add some clarity, I think you should make a couple changes:

> It's alright for `show.html`*`.erb`* to...
> `_stats.html`*`.erb` should not know... (also include the underscore prefix)

This will make it more clear that `stats.html` is actually a partial and that we're working in erb code. Thoughts?

stevegrossi:
Sorry I spaced out on this. Thanks for looking for (and finding) some of the context. Here's how it seems to me:

```ruby
# app/controllers/lessons_controller.rb
def show
  @lesson = Lesson.find...
end

# app/views/lessons/show/html.erb
render "lessons/overview/main"

# app/views/lessons/overview/_main.html.erb
render "lessons/overview/progress"

# app/views/lessons/overview/_progress.html.erb
@lesson.questions.each...
```

For one thing, it feels a little wrong that `lessons/overview/_progress` depends on `@lesson` being defined when it's 3 steps away in the dependency chain from the controller action that actually defines that instance variable.

But more practically, there's also a very similar `courses/overview/_progress` partial, and it would make sense to combine them into a `shared/overview/_progress` partial, but because both are dependent on an instance variable defined in the controller, there aren't a lot of great options:

- `(@lesson.defined? @lesson : @course).questions.each...` — yuck
- replacing `@lesson` and `@course` in the controller actions with something generic like `@assignable` that I could use in `shared/overview/_progress.html`, but that comes at the cost of clarity

Instead, if this dependency chain were written more explicitly:

```ruby
# app/controllers/lessons_controller.rb
def show
  @lesson = Lesson.find...
end

# app/views/lessons/show/html.erb
render "lessons/overview/main", lesson: @lesson

# app/views/lessons/overview/_main.html.erb
render "lessons/overview/progress", lesson: lesson

# app/views/lessons/overview/_progress.html.erb
lesson.questions.each...
```

Then it would be an easy refactor to pull out `shared/overview/_progress` and have it operate on any object passed in:

```ruby
# app/views/lessons/overview/_main.html.erb
render "shared/overview/progress", assignable: lesson

# app/views/courses/overview/_main.html.erb
render "shared/overview/progress", assignable: course

# app/views/shared/overview/_progress.html.erb
assignable.questions.each...
```

I hope this clarifies my thoughts a bit better. I don't think this is a major issue, but it does feel like a good practice to consider adopting.

(And good point @armilam, I'll update this!)

armilam:
Looks great!

PR #25, opened 2015-09-30
double quotes + close parens

chuckshulicker:
@lessonly/developers Been coming across a lot of methods without parens and it's making me eye-twitch a bit. Also added double-quotes whenever possible. In addition, I provided a few bad and good examples whenever possible so the guide is more digestable at a glance.

### DIFF
--- a/ruby-styleguide.md
+++ b/ruby-styleguide.md
@@ -13,44 +13,37 @@ Regarding coding style, we follow [Github's Ruby Style Guide](https://github.com
 These override either Github’s or Batsov’s styleguide where applicable:
 
 - Use double-quotes whenever possible.
-```ruby
-# bad
-  name = 'foo'
 
-# good
-  name = "foo"
-```
+      # bad
+      name = 'foo'
+
+      # good
+      name = "foo"
 - Close your parens when invoking a method.
-```ruby
-# bad
-  many_things.include? thing
 
-# good
-   many_things.include?(thing)
-```
+      # bad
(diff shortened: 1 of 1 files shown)


ruby-styleguide.md
```diff
@@ -12,23 +12,59 @@ Regarding coding style, we follow [Github's Ruby Style Guide](https://github.com
...
+  name = "foo"
+```
+- Close your parens when invoking a method.
+```ruby
+# bad
+  many_things.include? thing
+
+# good
+   many_things.include?(thing)
+```
```
stevegrossi: I love these examples, but they're not being rendered correctly by jekyll (which serves up our styleguide):

screen shot 2015-09-30 at 11 51 29 am

Unfortunately, our default jekyll install doesn't support Github-flavored markdown (we'd love to change that, just haven't found the time), so things like fenced code blocks (the triple-backtick) don't work. Instead you have to indent 4 spaces. And when code blocks are within a list element as they are here, they need to be indented 6 spaces (4 for the code block, and 2 for the list).

It's a bit of a pain, but I'd suggest pulling down this repo and running it locally with `$ jekyll serve` to confirm the formatting looks like you intend. That's how I got the screenshot above.

armilam:
Yeah, I'm not a fan of the parentheses-less method calls, either. We've already been changing single quotes to double quotes when we find them, too. This looks like a great addition!

armilam:
I can get behind this.

PR #31, opened 2016-01-13
Release Notes: Scheduled Assignments Summary

chuckshulicker:
@stevegrossi

### DIFF
--- a/release-notes.md
+++ b/release-notes.md
@@ -8,10 +8,6 @@ Look here for recent updates to the Lesson.ly platform.
 
 ### Coming soon
 
-- More good stuff!
-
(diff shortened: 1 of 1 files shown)


chuckshulicker:
>You may optionally set a future due date or a reassign on recurrence, so you can schedule an assignment once and never fuss with it again

I thought long and hard at how to explain recurrence and fell back on this...

stevegrossi:
Nice! I really like how you've explained this in non-technical terms. Nothing I would change about the text, but typically we'll put upcoming things under the "Coming Soon" heading (replacing `- more good stuff!`) and whoever deploys them will add the date in later. Why don't you make that change and then go ahead and merge this in?

PR #34, opened 2016-01-18
Use `floor` instead of `to_i` to round down

armilam:
As suggested by https://github.com/lessonly/lesson.ly/pull/1780#discussion_r50026273


stevegrossi:
Fantastic! Thanks for adding this. Onward into clarity!

PR #51, opened 2016-06-09
Add suggestions on when to use keyword arguments

### LINKED ISSUE #50: Ruby: suggest when to use named arguments
Discussion here: https://www.flowdock.com/app/lessonly/development/threads/LA6IwkfyFaJ-4HG51cT3e5lVF4J

But @stevegrossi and @baldwindavid both agree we should use named arguments whenever it isn't really obvious what the argument should be.

stevegrossi:
screen shot 2016-06-09 at 5 51 15 pm

Resolves #50. This has come up before, and [this Flowdock discussion](https://www.flowdock.com/app/lessonly/development/threads/LA6IwkfyFaJ-4HG51cT3e5lVF4J) prompted me to add it to the styleguide.

@baldwindavid, it sounds like you're on board. @lessonly/developers, thoughts?

Also:

- ignore the new .jekyll-metadata file, per Jekyll’s suggestion: https://github.com/jekyll/jekyll/issues/3595#issuecomment-83784543
- `bundle update rb-fsevent` because the version we had got yanked:

screen shot 2016-06-09 at 5 47 16 pm

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 _site
 .sass-cache
+.jekyll-metadata
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -102,7 +102,7 @@ GEM
       sawyer (~> 0.6.0, >= 0.5.3)
     public_suffix (1.5.3)
     rb-fsevent (0.9.7)
-    rb-inotify (0.9.6)
+    rb-inotify (0.9.7)
       ffi (>= 0.5.0)
     rdiscount (2.1.8)
     redcarpet (3.3.3)
@@ -127,4 +127,4 @@ DEPENDENCIES
   jekyll (~> 3.0)
 
 BUNDLED WITH
-   1.11.2
+   1.12.5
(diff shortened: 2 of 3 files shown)


baldwindavid:
@armilam I agree that the `can_view?` example is a lot more obvious as a candidate for keyword arguments. But more importantly, it brings up the question of whether the `notify_of_new_assignments` should use keywords or not. I think it would be helpful to use keyword arguments there, but can see how they might not be necessary. That gray area also pops up in a lot of cases where you may have only a single argument.

stevegrossi:
I like my example more than `can_view?` precisely because mine is closer to the border line between keyword args or not (and thus perhaps illustrates the gray area better), since with `can_view?` it's more obvious that keyword args are needed.

armilam:
Fair enough

baldwindavid:
Should we mention this preference even in the case of single argument methods or is that implicit enough?

stevegrossi:
I think it's probably implicit enough but then again I wrote it. I'm happy to add a single-arg example of when keyword args are helpful if anyone thinks it there may be ambiguity on that point.

baldwindavid:
I'm good with shipping it as is.

stevegrossi:
I'll wait till Monday to assign this to someone for merging, in case any final feedback rolls in.

armilam:
:shipit:

PR #60, opened 2017-02-24
Date and Time parsing advice

stevegrossi:
`Date.parse` (which ignores timezones and parses dates into the *server* timezone, not our *application's* timezone of Eastern, which we generally expect) has given us a lot of headaches in the past with time zone issues in the app and specs.

For that reason, we should always use the (unfortunately more verbose) `Time.zone.parse(string).to_date` when you really want a `Date` object, though the `Time` returned by `Time.zone.parse(string)` will often do as well.

We have [a note about time zones in our styleguide](http://lessonly.github.io/rails/#time-zones) but don't mention parsing. I want to fix that.

### DIFF
--- a/rails-style-guide.md
+++ b/rails-style-guide.md
@@ -89,3 +89,5 @@ end
 ### Time Zones
 
 Prefer `Time.current` and `Date.current` in place of `Time.now` and `Date.today`. The `.current` methods will properly take the `Time.zone` into account if set, which `.now` and `.today` do not, leading to inconsistent behavior.
+
+Likewise with parsing, avoid `Time.parse` and `Date.parse` (which also don’t account for timezones), and instead prefer `Time.zone.parse` if you need a `Time` object, and calling `to_date` on the result if you need a `Date`.


rreinhardt9:
In cases where we are concerned about timezone... would it make more sense to use a DateTime and therefore `DateTime.parse`? It kinda makes sense to me that a date would be timezone agnostic because it just represents a date, not any kind of time?

jacobsmith:
@stevegrossi is it worth making a helper class that handles this all for us? I think the advice is definitely good, but I could see a `Lessonly::TimeAndDate.parse(string).(time|date)` that handles the `Time.zone` things itself.

I think I tend to lean to helper objects really early, though, so definitely just want to see what your thoughts on that are.

jacobsmith:
@rreinhardt9 but the same Time can be many different Dates depending on where you are when we say "okay, NOW!"

Can we instead just abolish *all* timezones everywhere? Where's the PR for the world?

armilam:
> is it worth making a helper class that handles this all for us?

This sounds like a great idea to me. Perhaps in the future, that class can help us handle time zones for customers, too (if it needs to).

rreinhardt9:
I see what you're saying... that makes sense :+1:

stevegrossi:
Thanks for bringing up `Time` vs. `DateTime`, @rreinhardt9. I've always been fuzzy on the difference, mainly because it seems [there really isn't one](http://ruby-doc.org/stdlib-2.3.1/libdoc/date/rdoc/DateTime.html) unless you're dealing with pre-Gregorian dates (in which case `DateTime` is your friend):

screen shot 2017-02-24 at 1 04 32 pm

That said, re:

> It kinda makes sense to me that a date would be timezone agnostic 

I don't think it's conceptually possible for a date to be timezone agnostic. The date "February 24, 2017" means nothing—or rather, could mean 24 different things, one for each time zone—without the additional info of where, right?

Since `DateTime` does not seem to have a way to deal with timezones, I think that means `Time.zone.parse` is our only option here (short of wrapping it in our own helper, which I'd be happy to see addressed separately)

This seems to match what the Ruby docs eventually suggest:

screen shot 2017-02-24 at 1 15 00 pm

jacobsmith:
> or rather, could mean 24 different things

@stevegrossi if only it were even *that* easy - there's currently [39 different timezones](https://www.timeanddate.com/time/current-number-time-zones.html#zones)

jacobsmith:
Yeah, I think the only way to *truly* get timezone crap correct is to only use UTC and then just modify that based on the user's account, but everything is on one "absolute" time in the DB - and Lord help you when daylight saving time changes....

armilam:
I like the idea of a helper class, but this is a style guide, not a code change. I'm for merging this. All oppose?

baldwindavid:
@jacobsmith I'm pretty sure our database is all UTC, right?

baldwindavid:
And then `config.time_zone = "Eastern Time (US & Canada)"` just changes all of those to EST.

jacobsmith:
@baldwindavid I believe so - but only where Rails is concerned (I believe this is why most of our timestamps have `-5:00` on the end is that line)

PR #70, opened 2017-06-22
Internal JSON API style guide

### LINKED ISSUE #62: POSTing: Form data vs JSON
Do we prefer JSON now? There seem to be no downsides to JSON, and upsides like

- more concise
- more flexible
- consistency: we always return JSON, why not always accept it too?

stevegrossi:
Resolves #62 

screen shot 2017-06-22 at 11 05 00 am


json-style-guide.md
```diff
@@ -0,0 +1,11 @@
+---
+layout: page
+title: JSON API Style Guide
+permalink: /json/
+---
+
+Unless otherwise noted, these rules apply only to internal JSON endpoints, not necessarily [our public API](https://docs.lessonly.com).
+
+## Use JSON for request bodies.
```
stevegrossi: I think we should always use headings for styleguide rules, because Jekyll automatically generates fragment identifiers for them for easy linking in PRs and such:

screen shot 2017-06-22 at 11 02 58 am

andela-oesho:
LGTM 👍

json-style-guide.md
```diff
@@ -0,0 +1,11 @@
...
+
+Rails controllers handle JSON `params` as easily as form data, and JSON is more concise and flexible (especially when handling arrays and objects). Also, our endpoints always return JSON, so it’s more consistent also always to send it.
```
rreinhardt9: Makes sense to me; seems like a good best practice that is consistent with other well written APIs out there :+1:

json-style-guide.md
```diff
@@ -0,0 +1,11 @@
+---
+layout: page
+title: JSON API Style Guide
+permalink: /json/
+---
+
+Unless otherwise noted, these rules apply only to internal JSON endpoints, not necessarily [our public API](https://docs.lessonly.com).
```
rreinhardt9: What was the thinking behind excluding the public API? This hasn't really been a domain of our app that I've been working in recently... so sorry if this is a dumb question in this regard. Seems like it would be nice to standardize how we present the public API as well? Or is it mainly a backward compatibility thing...

stevegrossi: Yeah, mainly a backward-compatibility thing. With lots of customers depending on our public API, we're very limited in the changes we can make to it (excluding a major-version bump), even as we've learned better API practices. But we can be much more flexible for internal APIs.

But yeah, I agree that it would be great to standardize our public API practices as well, both the conventions we've stumbled into and the ones we'd like to adopt for the next major version.

PR #72, opened 2017-06-27
Give all rules a heading, and document this intention

stevegrossi:
So each rule can be linked to directly (e.g. in code reviews), because our theme adds `id` attributes with fragment identifiers to all headings:

screen shot 2017-06-26 at 8 01 09 pm

Also, the way we'd been doing rules before (as list items) requires code blocks to be indented which is awkward and difficult when using Github's online editor. We want to make it as easy as possible to add or modify this styleguide.

## Before

screen shot 2017-06-26 at 8 16 33 pm

## After

screen shot 2017-06-26 at 8 16 27 pm

## Testing Notes

Unless your brain has a markdown parser, you'll probably want to test this one locally:

```
$ cd lessonly.github.io
$ git checkout stevegrossi/use-headings-for-rules
$ bundle exec jekyll serve
```

and check out http://localhost:4000


meta-style-guide.md
```diff
@@ -0,0 +1,11 @@
...
+layout: page
+title: Meta Style Guide
+permalink: /meta/
+---
+
+This is the style guide page about the style guide. Deep, right?
+
+## Every rule should start with a heading.
+
+This is so that each rule gets a unique URL, since [our style guide theme](https://github.com/jekyll/minima) adds an `id` attribute to each heading which can be [linked to directly](/meta/#every-rule-should-start-with-a-heading).
```
stevegrossi: So meta!

rails-style-guide.md
```diff
@@ -8,57 +8,63 @@ While we don’t follow it explicitly, the [community Rails Style Guide](https:/
 
 ### Dealing with SQL
 
-- Prefer ActiveRecord methods to raw SQL
-  
-  ```ruby
-  # not so good
-  Model.where("published_at = ? AND title IS NOT NULL", Date.yesterday)
+#### Prefer ActiveRecord methods to raw SQL
```
stevegrossi: There are no content changes in the rest of this PR, only un-indenting and heading-ifying.

rreinhardt9:
Seems like there isn't any other input from the team on this, so I'll go ahead and merge. Thanks for your work on this @stevegrossi !!!! This will be nice :-)

PR #83, opened 2017-10-19
Always specify attributes that a test depends on

### LINKED ISSUE #33: Test guidelines: avoid relying on default/unspecified values
If a spec depends on something being `nil` and it's `nil` by default, it should still specify that fact both for obviousness and so if someone changes the default value, the spec will continue testing what it means to.

``` ruby
# Not so good:
lesson = build(:lesson)
expect(lesson.description).to be nil

# So good!
lesson = build(:lesson, description: nil)
expect(lesson.description).to be nil
```

stevegrossi:
Resolves #33

### DIFF
--- a/_styleguide/testing.md
+++ b/_styleguide/testing.md
@@ -3,11 +3,9 @@ layout: page
 title: Testing Style Guide
 ---
 
-## Rule #1
+### Rule #1
 
 Write tests for your code, please.
-
-## House Rules
 
 ### In RSpec tests, denote instance methods with "#" and class methods with "::"
 
@@ -61,8 +59,6 @@ click_button 'Save'
 expect(User.find_by(name: 'John Doe').name).to eq 'John Doe' # This line hits the DB directly
 ```
 
-## Guidelines
-
 ### Don't stub methods on the class being tested.
 
 [Read why](https://robots.thoughtbot.com/don-t-stub-the-system-under-test).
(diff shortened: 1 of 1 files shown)


_styleguide/testing.md
```diff
@@ -61,8 +59,6 @@ click_button 'Save'
 expect(User.find_by(name: 'John Doe').name).to eq 'John Doe' # This line hits the DB directly
 ```
 
-## Guidelines
-
```
stevegrossi: I got rid of the `## House Rules` and `## Guidelines` headings because they don't make sense and weren't being followed anyway.

stevegrossi:
For your consideration, @lessonly/developers

_styleguide/testing.md
```diff
@@ -87,6 +83,25 @@ before :each do
...
+
+``` ruby
+# Not so good: breaks unnecessarily if we add a default description
+it "is invalid without a description" do
+  lesson = build(:lesson)
+  expect(lesson.description).to be nil
+end
+
+# So good!
+it "is invalid without a description" do
```
baldwindavid: I agree with the spirit of this guidance. Should the specific example `expect` validation of `false/true` as opposed to an expectation of the `description`?

stevegrossi: Whoops, bad copypasta 🍝 on my part!

baldwindavid:
Thanks for adding! I went ahead and merged it.

PR #84, opened 2017-10-25
Update our style guide around Service objects

### LINKED ISSUE #55: Document service objects
Document #fail! and failure states for service objects

stevegrossi:
Resolves #55 

## Why?

- Our current styleguide is out-of-date! It neglects to mention:
  - our `Service` base class
  - its `fail!` and `success?` methods
  - Our `rails g service` custom generator
  - How to return data from a service (i.e. as a public instance method or `attr_reader`)

## What?

See the diff attached, but if you just want to read the end result:

screen shot 2017-10-25 at 4 52 54 pm

the rest:

screen shot 2017-10-25 at 4 53 09 pm

## ~~Testing~~ Reading Notes

- Is this clear and makes sense, especially to a new team member?
- Is the example correct, and a good illustration of how we use Services?


stevegrossi:
Thoughts, @lessonly/developers?

_styleguide/ruby.md
```diff
@@ -180,51 +180,59 @@ The latter introduces load order dependencies (`MyModule` must already be loaded
...
+We use the service object pattern to encapsulate complex or repeated logic within self-contained objects. Instead of a 20-line controller action, we’ll call `DoSomethingComplicated.perform(args)` in the controller and move all of the complex logic into the `DoSomethingComplicated` class. This not only helps organize and isolate complex code, but also makes it easier to test. It also reduces duplication of logic between our web app and our API, for example, since both can rely on the same service classes.
 
-Service objects are called with `ServiceName.perform`. They should be named imperatively after the actions they perform (e.g. `AssignUser`, `CreateGroup`).
+- New service classes should descend from our `Service` base class.
+- Services should be named imperatively after the actions they perform (e.g. `AssignUser`, `CreateGroup`).
```
wernull: Should we say anything about [verb][noun] for naming?

_styleguide/ruby.md
```diff
@@ -180,51 +180,59 @@ The latter introduces load order dependencies (`MyModule` must already be loaded
...
+- Services are called with the `::perform` class method, e.g. `ServiceName.perform(args)`, which is defined on `Service`. It will call `#perform` on a new instance of the service class, which is where we define the service’s logic.
+- `::perform` always returns the instance of the service, which can be queried for status and results.
+- Service instances return `true` for `#success?` by default. Call `fail!` at any point to have them respond to `#success?` with `false`.
```
joshmfrankel: > Call `fail!` at any point to have them respond to `#success?` with `false`.

Somewhat related thought. Currently, we use things like `return fail! if some_condition?` throughout our services. The only place this doesn't work is within the initialize method. I'd love to make this available in our services so that early returns work in Service classes like they do everywhere else. Just trying to think through any gotchas a new developer might have.

_styleguide/ruby.md
```diff
@@ -180,51 +180,59 @@ The latter introduces load order dependencies (`MyModule` must already be loaded
...
+    render :new, error: "There was a problem."
   end
 end
 ```
 
-While perhaps overkill in this simple case, it's valuable to be able to scan a service object's `#perform` method and see, on each line, all of its responsibilities.
+Note that you can generate new services with:
+
+```sh
+$ rails generate service DoSomething
```
joshmfrankel: :heart_eyes: :revolving_hearts: :heart_eyes:

_styleguide/ruby.md
stevegrossi: > The only place this doesn't work is within the initialize method.

Good point. I'd like to consider that separate from this styleguide PR since it'll likely require changes to our `Service` base class—but I'll be interested in ideas around this!

_styleguide/ruby.md
case-eee: 👍

armilam: I had no idea

stevegrossi:
I made a small addition to mention the [Verb][Noun] convention, per @wernull's suggestion!

wernull:
Thanks for adding that. I'll hit the merge button!

PR #85, opened 2017-11-02
Remove suggestion to query the database in feature specs

stevegrossi:
Our feature tests suffer from nondeterministic failures (flakes). [This is a great read](https://bibwild.wordpress.com/2016/02/18/struggling-towards-reliable-capybara-javascript-testing/) on all the things that may be going wrong, one of which is that Capybara does not intend to fully support issuing database queries from feature tests. As the creator says:

> Capybara is written to best support use cases where you only test the UI, not the back-end.

So this PR removes the suggestion that we do that, hopefully to reduce the potential for flakes.

### DIFF
--- a/_styleguide/testing.md
+++ b/_styleguide/testing.md
@@ -48,6 +48,26 @@ feature "Custom user fields" do
 end
 ```
 
+### In feature specs, test the UI, not the database.
+
+In the words of Capybara’s creator, “Capybara is written to best support use cases where you only test the UI, not the back-end.” ([source](https://bibwild.wordpress.com/2016/02/18/struggling-towards-reliable-capybara-javascript-testing/)). Calling out to the database in feature specs can cause threading issues. For example:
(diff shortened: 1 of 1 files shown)


_styleguide/testing.md
```diff
@@ -48,17 +48,6 @@ feature "Custom user fields" do
 end
 ```
 
-We also have the ability to query the database directly from the feature specs, so a feature test can change something via the UI, but verify it with actual database calls.
```
wernull: What if we kept this in as a warning of what not to do. "We have the ability, but don't"

stevegrossi: Good point 👍

stevegrossi: Great suggestion, @wernull—added!

stevegrossi:
If you're cool with this, @wernull, can ya spare a merge?

PR #86, opened 2018-01-04
Add documentation to style guide

rreinhardt9:
To be updated:

- [x] Add a code example of using yard with markdown
- [x] Link to the yard docs
- [x] Add verbiage that explains that we are using markdown in the comments instead of rdoc
- [x] Remove the line about public methods "should" be documented as we are leaving that up to the dev


joshmfrankel:
I think I hit all of our change points listed above. Take a peek and see if we need to add anything else or expand on the example.

rreinhardt9:
Nice!!! I love the updates @joshmfrankel thanks for taking that on! I think if nothing else, it's a great start and we'll learn as we go :100:

rreinhardt9:
Merged! Thanks Y'all!

Issue #67, opened 2017-06-14
Document our use of the policy pattern and pundit


stevegrossi:
https://github.com/elabs/pundit

stevegrossi:
- Policy scopes are responsible for answering "What records should a user even know exist?"
- Policy predicate methods assume the user can know about the record, and are responsible for answering "Can this user take some action on a record?"

For example, a common pattern in controllers is to find a record through a policy scope and then `authorize` a particular action on it:

```ruby
def show
  @lesson = policy_scope(Lesson).find(params[:id])
  authorize @lesson # implicitly checks the show? predicate method because it's the current action
end
```

This way, the `LessonPolicy#show?` predicate method doesn't need to check that the lesson belongs to the same company as the user: we can assume that's already been done.

stevegrossi:
We should include plenty of examples (do policies need their own style guide page?), like how to dynamically instantiate polymorphic policies:

```ruby
# app/controllers/taggings_controller.rb
def create
  taggable = policy_scope!(params[:taggable_type].safe_constantize).find(params[:taggable_id])
end
```

Should we whitelist `taggable_type` in this case? Probably, but how?

stevegrossi:
> For example, a common pattern in controllers is to find a record through a policy scope and then authorize a particular action on it

[We don't want to do this anymore, Steve!](https://lessonly.slack.com/files/U048AJQRZ/F7L9DGW8G/Untitled.rb)

```ruby
# Instead of this (how we usually do things)
def update
  # checks whether you can know the lesson exists:
  @lesson = policy_scope(Lesson).find(params[:id])
  # checks whether your role lets you update it:
  authorize @lesson
end

# We propose this instead
def update
  @lesson = Lesson.find(params[:id])
  # checks whether you can know the Lesson exists
  # *and* that your role has permission to update @lesson
  authorize @lesson
end
```

stevegrossi:
@baldwindavid [started thinking](https://github.com/lessonly/lessonly/pull/4414#discussion_r155058226) about what the best convention for calling policy scopes *within* policies (when needed) would be:

```ruby
def update?
  Scope.new(user, Tag).include?(record)
end
```

I like that, though I would add that it could be a bit more concise thanks to a shortcut in `PolicyHelpers`:

```ruby
def update?
  policy_scope!(Tag).include?(record)
end
```

and potentially even shorter with new code, though perhaps at the cost of clarity:

```ruby
def update?
  in_scope?(record)
end

[...]

PR #92, opened 2018-02-13
Documentation for Policies, Support for subdirectory styleguide topics

joshmfrankel:
Implements #67


_styleguide/ruby/policy.md
```diff
@@ -0,0 +1,82 @@
...
+We use the policy pattern implemented by the [Pundit gem][1] to create a flexible authorization system.
+
+## Overview
+
+* Policies are located within `app/policies`
+* Each model has a policy associated with it
+* Policy predicate methods (`#create?`, `#update?`, `#destroy?`, etc) determine if the `current_user` can perform the given action for the record
+* Policy scopes are responsible for answering "What records should a user even know exist?"
+
+## Policies
```
joshmfrankel: - [ ] TODO: Needs details and examples

_styleguide/ruby/policy.md
```diff
@@ -0,0 +1,82 @@
...
+
+### Load accessible records through the Policy Scope
+
+```ruby
+# Only loads records which the current_user's role has access to
+# from the Policy scope
+def filtered_collection
+  collection = policy_scope(user.criteria)
+end
+```
```
joshmfrankel: How to dynamically instantiate polymorphic policies within controllers?

_styleguide/ruby/policy.md
```diff
@@ -0,0 +1,82 @@
...
+* Policy scopes are responsible for answering "What records should a user even know exist?"
+
+## Policies
+
+## Controllers
+
+* Controllers should include the `Policies` concern to ensure all actions are verified.
+
+* **Only when necessary**: Controller verification checks may be skipped by using the following: `skip_after_action :verify_authorized`
+
```
joshmfrankel: - [ ] Todo: More detail on best practices
- [ ] authorize_with_policy method and undocumented namespacing https://github.com/varvet/pundit/pull/392/files

_styleguide/ruby/policy.md
```diff
@@ -0,0 +1,82 @@
...
+### Load accessible records through the Policy Scope
+
+```ruby
+# Only loads records which the current_user's role has access to
+# from the Policy scope
+def filtered_collection
+  collection = policy_scope(user.criteria)
+end
+```
+
```
joshmfrankel: Controller specs and handle_unauthorized

_config.yml
```diff
@@ -18,7 +18,10 @@ permalink: /:title/
 collections:
   styleguide:
     output: true
-    permalink: /styleguide/:path/
+    permalink: /:collection/:path
+  ruby_styleguide:
+    output: true
+    permalink: /styleguide/ruby/:path
```
joshmfrankel: I wanted specific ruby patterns and/or sections to be able to be broken out by folder. Alas the above is the best I could come up with. Jekyll doesn't play nice with subdirectories for collections.

My initial approach
```
styleguide/  # Contains top level docs with links to specific sections
styleguide/ruby/ # Contains specific ruby docs (most likely patterns)
```

What I did above:

```
styleguide/  # Contains top level docs with links to specific sections
ruby_styleguide/ # Contains specific ruby docs (most likely patterns)
```

joshmfrankel:
TODO:
- [ ] How to dynamically instantiate polymorphic policies within controllers?
- [ ] Writing specs for policies? How with examples
- [ ]  authorize_with_policy method and undocumented namespacing https://github.com/varvet/pundit/pull/392/files
- [ ] Controller specs and handle_unauthorized

_config.yml
joshmfrankel: actually we might be able to use

`collections_dir: styleguide`

and then fill out individual folders within /styleguide/

_styleguide/ruby/policy.md
```diff
@@ -0,0 +1,82 @@
+---
+layout: page
+title: "Ruby: Policy Pattern"
```
case-eee: @joshmfrankel This is a great start 🎉 I'm going to work on adding to this 👍

_config.yml
joshmfrankel: I tried to make this work again and alas we need jekyll 3.7.2 installed. Unfortunately, github-pages gem 177 (newest version) requires jekyll 3.6.2 exactly which means we are stuck :-(

_styleguide/ruby.md
```diff
@@ -1,8 +1,27 @@
...
+There are specific architecture patterns that we utilize within the Lessonly application. Below are page specfic resources detailing when to use and how to implement the respective patterns.
+
+<ul class="post-list">
+  {% for pattern in site.ruby_styleguide %}
+    <li>
+      <h2>
+        <a class="post-link" href="{{ pattern.url | prepend: site.baseurl }}">{{ pattern.title }}</a>
+      </h2>
+    </li>
+  {% endfor %}
```
joshmfrankel: This is the styleguide overview for ruby file which now contains a table of contents that points to all markdown files within _ruby_styleguide. This should help us break out specific patterns into their own sections thereby reducing how long this styleguide has become.

_config.yml
stevegrossi: In the past, I've only used Jekyll "collections" are for things like blog posts that need special handling like sorting and filtering. By default, Jekyll by default treats all files and directories not prefixed with an underscore as pages. Could we use plain ol' pages for these? This seems to accomplish what we want:

```txt
styleguide/ (note it's no longer /_styleguide/)
  javascript.md (/styleguide/javascript/)
  ...
  ruby/
    index.md (/styleguide/ruby/)
    services.md (/styleguide/ruby/services/)
    ...etc
```

joshmfrankel: This should work well for organization great idea reusing existing Jekyll functionality! Question I have on this approach is can we still iterate through all the pages like we are doing for the lists of links? Basically we'll still need a way on styleguide.md and ruby.md of showing their specific contents:

{% for pattern in site.ruby_styleguide %}
{% for pattern in site.styleguide %}

I'm open to thoughts/ideas as I would love a more organized folder structure as you showed above :+1:

stevegrossi: Ah, iteration! I think that's one of the other benefits of collections. It may be possible to iterate through `site.pages` and filter pages by category or some other metadata, or we could just hard-code links to the Ruby sub-pages on `/styleguide/ruby/` (a bit more work, but they may not change very often and I like the simplicity of pages).

Your collection/category-based approach sounds like it'll do what we need though!

joshmfrankel: @stevegrossi I actually ran with what you mentioned by using pages. For posts there isn't a way to iterate through categories with what GIthub-pages supports (without using a build pipeline like TravisCI).

So by using pages and a bit of liquid templating magic I was able to construct a way for us to iterate through subpages for pages like Ruby. Check it out locally

/styleguide/
/styleguide/ruby/
/styleguide/<insert pattern name here>

_config.yml
```diff
@@ -15,14 +15,6 @@ kramdown:
...
 permalink: /:title/
 
-collections:
-  styleguide:
-    output: true
-    permalink: /:collection/:path
-  ruby_styleguide:
-    output: true
-    permalink: /styleguide/ruby/:path
-
```
joshmfrankel: Removed! Collections are only so good in 3.6.2 and we need the newest version in order to have more granular control here. Because of this I've switched our styleguide to a page iterating setup. See below

_includes/subpage_listing.html
```diff
@@ -0,0 +1,12 @@
...
+    {% assign current_path = subpage.path | prepend: "/" %}
+    {% if current_path contains page.url %}
+      <li>
+        <h2>
+          <a class="post-link" href="{{ subpage.url | prepend: site.baseurl }}">{{ subpage.title }}</a>
+        </h2>
+      </li>
+    {% endif %}
+  {% endfor %}
+</ul>
```
joshmfrankel: An include that works on top level pages, like ruby.md, to iterate through all files in a similarly named directory `/styleguide/ruby/`. We should be able to reuse this in the future for at least one more level of subdirectory. Not sure if it will work for 2 or more

styleguide/index.md
```diff
@@ -16,11 +16,20 @@ The app contains an [`.editorconfig`](https://github.com/lessonly/lessonly/blob/
...
+      and dir_depth == 2
+      and page.path != "styleguide/index.md" %}
+      <li>
+        <h2>
+          <a class="post-link" href="{{ page.url | prepend: site.baseurl }}">{{ page.title }}</a>
+        </h2>
+      </li>
+    {% endif %}
   {% endfor %}
 </ul>
```
joshmfrankel: I had to get a bit creative with the above approach. liquid templates don't have regular expressions so we have to count the directories to tell the depth within the /styleguide folder. This is without re-outputting the current index.md file

styleguide/ruby.md
```diff
@@ -0,0 +1,191 @@
+---
+layout: page
+title: Ruby Style Guide
+---
+
+## Patterns / Concepts
+
+There are specific architectural patterns and concepts that we utilize within the Lessonly application. Below are page specfic resources detailing when to use and how to implement the respective theory.
+
+{% include subpage_listing.html %}
```
joshmfrankel: Ahhhh, yeah! This outputs every md file within the directory` /styleguide/ruby/`

joshmfrankel:
I hacked around for a bit and made our styleguide more organized based on the suggestion from @stevegrossi. We now can use subfolders to organize subsequent styleguides into sections with a handy include file :tada:

styleguide/ruby/policy.md
```diff
@@ -0,0 +1,180 @@
+---
```
joshmfrankel: This file is the bulk of the new styleguide changes

styleguide/ruby/policy.md
```diff
@@ -0,0 +1,180 @@
...
+# Use discretion when basing actions off one another
+def destroy?
+  create? # Just because a user can create doesn't mean they can destroy records
+end
+```
+
+#### Helpers
+
+* We have several helpers for determining the current user's role. These can be
+used within scopes to great affect.
```
joshmfrankel: affect? effect? I never get these right :confused:

stevegrossi: Apparently, the RAVEN's got your back: https://www.vocabulary.com/articles/chooseyourwords/affect-effect/

> Remember: affect verb, effect noun

Not the best mnemonic, but it might help!

case-eee: I always use "impact" and avoid these two words 😬

joshmfrankel:
@stevegrossi and @case-eee, I think this one is good to go for review/merge :)

PR #96, opened 2018-07-27
Add documentation around testing to style guide

haleymnatz:
Resolves [[ch13791]](https://app.clubhouse.io/lessonly/story/13791)

Adds helpful documentation to testing portion of style guide.

### DIFF
--- a/styleguide/testing.md
+++ b/styleguide/testing.md
@@ -110,6 +110,22 @@ it "is invalid without a description" do
 end
 ```
 
+### Use the `request_setup` helper in your controller specs.
+
+Controller specs won't work without it! Its definition can be found in `spec/support/controller_helper.rb/`. It accepts `request`, `controller`, `current_user`, and `current_company` as arguments, but you can substitute other objects for those as needed.
+
+```ruby
+context "when no user is logged in" do
+  let(:company) { create(:company) }
+
+  before(:each) { request_setup(request, controller, nil, company) }
+
+  it "redirects to the login page" do
+    # some code...
+  end
+end
+```
+
 
 ## RSpec Syntax
 
--- a/styleguide/testing.md
+++ b/styleguide/testing.md
@@ -113,6 +113,34 @@ end
 
 ## RSpec Syntax
 
+- Use `specify` instead of `it` when it improves readability. This is common in our jobs specs, where `context` is used to refer to subclasses, and our model specs, where we test validations.
+
+Example:
+
+```ruby
+# reads weirdly...
+describe DoStuffJob do
+
+  context DoStuffJob::Subclass do
+
+    it "method does this thing" do
+      # some code
+    end
+  end
+end
+
+# reads so smoothly!
+describe DoStuffJob do
+
+  context DoStuffJob::Subclass do
+
+    specify "method does this thing" do
+      # some code
+    end
+  end
+end
+```
+
 - Include a blank line around `describe`/`feature` blocks, `it`/`scenario` blocks, `before`/`background` blocks, and `context` blocks.
 - Within `it`/`scenario` blocks, separate setup and expectations with a blank line.
 - Do not include a blank line above `let` blocks, in order to visually group them with their appropriate `describe`/`feature` or `context` block.


styleguide/testing.md
```diff
@@ -110,8 +110,96 @@ it "is invalid without a description" do
...
+
+  it "redirects to the login page" do
+    # some code...
+  end
+end
+```
+
+### JS Helpers
+
+In `spec/support/javascript_helper.rb`, there are a number of methods to help with unique JS issues that often come up in feature specs. Many of these methods are titled `wait_and_[do_something]` and address timing issues that come up when a request is made before a JS request has fully processed.
```
rreinhardt9: This is true, but I would note that the preferred way to way for JS to finish would be to watch for a change on the page using Capybara rather that using one of these slow "wait_for" helpers. For example, instead of wait_for you could `expect(page).to have_content("It's finished!")` and that would wait... same if the change is it's going away `expect(page).not_to have_content("It's finished!")`

styleguide/testing.md
```diff
@@ -110,8 +110,96 @@ it "is invalid without a description" do
...
+### Use `and_call_original` to call the actual implementation of a mocked method
+
+Why might you need this?
+
+- To "unmock" a method that was previously mocked
+- To set a default response that can be overridden with specific arguments
+- To make sure that mocked services called in job specs actually carry out their task
+
+```ruby
+What example should go here?
```
armilam:
```rb
# Stub getting a value from the session, but make sure calls to get unrelated values from the session remain unstubbed
allow(controller.session).to receive(:[]).and_call_original
allow(controller.session).to receive(:[]).with("gradebook_user_ids") { [user.id] }
```

styleguide/testing.md
```diff
@@ -110,8 +110,96 @@ it "is invalid without a description" do
...
+  This not only speeds up the tests but also enhances separation of concerns. If your code is written in a way that makes writing tests like this difficult, that might be a sign that your code could be refactored.
+
+#### If you are adding tests to an existing file, check for code that can be reused.
+
+  Creating a new `user` when the same type of `user` has already been created in the file creates an unnecessary test object to add to memory.
+
+
+### Use the `request_setup` helper in your controller specs.
+
+Controller specs won't work without it! Its definition can be found in `spec/support/controller_helper.rb/`. It accepts `request`, `controller`, `current_user`, and `current_company` as arguments, but you can substitute other objects for those as needed.
```
unixmonkey: Super minor thing, but the trailing slash after `.rb/` makes it look like this could be a directory instead of file.

styleguide/testing.md
```diff
@@ -110,8 +110,96 @@ it "is invalid without a description" do
...
+  > build_stubbed ... instantiates and assigns attributes just like build ... It makes objects look like they’ve been persisted, creates associations with the build_stubbed strategy ... and stubs out a handful of methods that interact with the database and raises if you call them. This leads to much faster tests and reduces your test dependency on a database.
+
+#### Stub and mock external services and classes.
+
+  This not only speeds up the tests but also enhances separation of concerns. If your code is written in a way that makes writing tests like this difficult, that might be a sign that your code could be refactored.
+
+#### If you are adding tests to an existing file, check for code that can be reused.
+
+  Creating a new `user` when the same type of `user` has already been created in the file creates an unnecessary test object to add to memory.
+
```
unixmonkey: Again, super minor, but inconsistent spacing between sections. Is it one carriage return or two?

styleguide/testing.md
unixmonkey: Yes `javascript_helper.rb` has several methods with a `sleep` in them, and likely overall negatively contribute to test run times.

We should prefer to use Capybara methods that have built in change detection where possible. [Here's a nice concise guide on methods that wait vs ones that don't, and how to adjust](http://stefan.haflidason.com/testing-with-rails-and-capybara-methods-that-wait-method-that-wont/), and where that doesn't work well (or is flaky), then look to `javascript_helper.rb`

haleymnatz: @rreinhardt9 @unixmonkey Thanks for that correction! I'll definitely include a link to that article. Are there any other things you think would be useful to know about from `javascript_helper.rb`? Do you think I should keep the above point about the file, or should I move the focus to something like "Prefer to use Capaybara methods with built in change detection over JS helper methods" ?

styleguide/testing.md
haleymnatz: Whoops! That's definitely a typo. Thanks for being so thorough!

styleguide/testing.md
haleymnatz: I think it's just one, although there were a few sections that had two when I started working in the file. But I'll go ahead and fix that!

styleguide/testing.md
unixmonkey: I like what you put here, but I might preface the above paragraph with one like this:

> Tests that execute JavaScript need to wait for a virtual browser to boot up, navigate to a URL, then load our JS before we can fully start testing it. There is a delay between clicking something and an action taking place, or an Ajax call to return data to render a component.[Here's a handy guide on what Capybara methods wait for an element or selector to appear, and which will attempt to run immediately](http://stefan.haflidason.com/testing-with-rails-and-capybara-methods-that-wait-method-that-wont/). Prefer to use Capybara methods that wait for an element to appear before performing
```

File: styleguide/testing.md [hunk 5]
```diff
@@ -126,6 +126,10 @@ context "when no user is logged in" do
 end
 ```
 
+### Something about how shared contexts are used and/or helpful
```
gardengeek: Would it also be helpful to include something about using `shared_examples` for reusing common tests? ie. `shared_contexts` is common setup code that can be used to prepare a test case, while `shared_examples` are tests written in a way that they can be run in multiple settings; extracting common behavior between objects.

rreinhardt9: ooh good call... one of the best uses I've seen for shared contexts is when you have duck type that are playing a certain role and you expect them to have a particular interface. In ruby, we don't declare interfaces... but a great way to test/ document that the object under test is a particular duck type is by having a shared test that is mixed in and test that! A good example in the app of this is `LessonContent`. In things that are lesson content (like a scorm for example) we `include LessonContent` and then in the tests we assert `it_behaves_like "lesson content"`

styleguide/testing.md
```diff
@@ -110,8 +110,101 @@ it "is invalid without a description" do
 end
 ```
 
+### Write your specs with performance in mind.
```
caryssaperez: Can also mention feature specs and using 
```ruby
expect(page).not_to have_content
``` 
to 
```ruby
expect(page.has_no_content?).to eq(true)
```

so we can avoid instances where Capybara might wait for the full `default_max_wait_time`. Especially important too since Ross upped it to 8 seconds.

More matchers can be found [here](https://www.rubydoc.info/gems/capybara/0.4.0/Capybara/Node/Matchers)

styleguide/testing.md
```diff
@@ -110,8 +110,101 @@ it "is invalid without a description" do
...
+
+Do these things to minimize contributions to the slowness of our test suite:
```
stevegrossi: This may just be taste, but "the slowness of our test suite" sounds kind of "glass half empty" to me, especially in light of recent contributions to making it faster through `rspec-retry` and `parallel-tests`. I might suggest wording like "Here's how you can help keep our test suite running fast:"

styleguide/testing.md
```diff
@@ -110,8 +110,101 @@ it "is invalid without a description" do
...
+
+#### Default to using `build_stubbed` instead of `create`
+
+  In general, it works in most places where you just need objects to exist without them interacting with the database (like for creating associations).
+
+  From [this post from Factory Bot](https://robots.thoughtbot.com/use-factory-girls-build-stubbed-for-a-faster-test):
+
+  > build_stubbed ... instantiates and assigns attributes just like build ... It makes objects look like they’ve been persisted, creates associations with the build_stubbed strategy ... and stubs out a handful of methods that interact with the database and raises if you call them. This leads to much faster tests and reduces your test dependency on a database.
```
stevegrossi: Great quote! Do you think keeping specific code marked as such with backticks would aid readability?

> `build_stubbed` ... instantiates and assigns attributes just like `build` ... It makes objects look like they’ve been persisted, creates associations with the `build_stubbed` strategy ... and stubs out a handful of methods that interact with the database and raises if you call them. This leads to much faster tests and reduces your test dependency on a database.

instead of

> build_stubbed ... instantiates and assigns attributes just like build ... It makes objects look like they’ve been persisted, creates associations with the build_stubbed strategy ... and stubs out a handful of methods that interact with the database and raises if you call them. This leads to much faster tests and reduces your test dependency on a database.

styleguide/testing.md
```diff
@@ -110,8 +110,101 @@ it "is invalid without a description" do
 end
 ```
 
+### Write your specs with performance in mind.
+
+Do these things to minimize contributions to the slowness of our test suite:
+
+#### Default to using `build_stubbed` instead of `create`
+
+  In general, it works in most places where you just need objects to exist without them interacting with the database (like for creating associations).
```
stevegrossi: It looks like new paragraphs are indented with two spaces in the markdown, however Markdown will ignore these spaces:

screen shot 2018-09-04 at 10 06 15 am

In the interest of avoiding unnecessary text/code, what do you think of not indenting paragraphs in the raw Markdown, as they're already demarcated by newlines?

```markdown
#### Default to using `build_stubbed` instead of `create`

In general, it works in...
```

styleguide/testing.md
```diff
@@ -110,8 +110,101 @@ it "is invalid without a description" do
...
+
+  From [this post from Factory Bot](https://robots.thoughtbot.com/use-factory-girls-build-stubbed-for-a-faster-test):
+
+  > build_stubbed ... instantiates and assigns attributes just like build ... It makes objects look like they’ve been persisted, creates associations with the build_stubbed strategy ... and stubs out a handful of methods that interact with the database and raises if you call them. This leads to much faster tests and reduces your test dependency on a database.
+
+#### Stub and mock external services and classes.
+
+  This not only speeds up the tests but also enhances separation of concerns. If your code is written in a way that makes writing tests like this difficult, that might be a sign that your code could be refactored.
```
stevegrossi: While I think this is a good guideline, I've definitely taken this one too far myself, leading to code that was impossible to refactor without changing the tests. (And [I'm not the only one](https://8thlight.com/blog/eric-smith/2010/02/08/stubbing-new-considered-harmful.html).) I'd love to have some guidance (or a link!) here on how "code could be refactored" to avoid this guideline leading us down the wrong path, but we can always add that later on!

styleguide/testing.md
```diff
@@ -110,8 +110,101 @@ it "is invalid without a description" do
...
+
+  it "redirects to the login page" do
+    # some code...
+  end
+end
+```
+
+### JS Helpers
+
+In feature specs that use JavaScript to return data or render a component, there is a delay between clicking something and an action taking place. Prefer to use Capybara methods that wait for an element to appear before acting on them with other methods that don't wait. Here's a [handy guide](http://stefan.haflidason.com/testing-with-rails-and-capybara-methods-that-wait-method-that-wont/) on which Capybara methods wait for an element or selector to appear, and which will attempt to run immediately.
```
stevegrossi: That link is awesome 💯 —I wish I'd discovered it sooner!

styleguide/testing.md
```diff
@@ -110,8 +110,101 @@ it "is invalid without a description" do
...
+
+However, sometimes that doesn't work well (or can be flaky). If you are running into that kind of issue, check out /spec/support/javascript_helpers.rb.
```
stevegrossi: It looks to me like the file is `/spec/support/javascript_helper.rb` (no plural "s" in helpers.) Also, elsewhere it looks like you demarcate filenames in backticks, e.g.

> Shared contexts are defined in `spec/support/shared_contexts.rb`

which I like because it makes them easier to copy into a terminal! Should we do that here as well?

styleguide/testing.md
```diff
@@ -110,8 +110,101 @@ it "is invalid without a description" do
...
+# reads so smoothly!
+describe DoStuffJob do
+
+  context DoStuffJob::Subclass do
+
+    specify "method does this thing" do
+      # some code
+    end
+  end
+end
```
stevegrossi: While I agree in principle that `specify` can sometimes read better than `it`, I don't think this example makes the most compelling case for that. If we had more than one test for a single method, using `specify` seems like it would introduce duplication of the "method" in question, leading to trouble if the method is ever renamed:

```ruby
context DoStuffJob::Subclass do
  specify "method does this thing" do
    # some code
  end

   specify "method does this other thing" do
    # some code
  end
end
```

compared with an additional block for the method:

```ruby
describe DoStuffJob::Subclass do

  describe "#method" do

    it "does this thing" do
      # some code
    end

     it "does this other thing" do
      # some code
    end
  end
end
```

To be fair, I do see examples of this in our codebase—e.g. `spec/jobs/smart_groups/recalculate_memberships_job_spec.rb` the duplicating `specify "#perform does...`—I'm just not sure it serves us to embrace that over more concise and DRY alternatives, would you agree?

If so, the guideline I would suggest here would be something like this:

> Prefer the more-concise `it` when declaring test scenarios about specific things (e.g. classes, methods), but consider `specify` when more grammatical flexibility is required.
> 
> ```ruby
> describe CookiePolicy do
>   # Ehh...
>   it "admins are permitted eat cookies" do
>     # ...
>   end
> 
>   # That's better!
>   specify "admins are permitted eat cookies" do
>     # ...
>   end
> end
> ```
> 
> Though keep in mind that specs using `specify` can typically be written more concisely using `it`, e.g.
> 
> ```ruby
> describe CookiePolicy do
>   specify "admins are permitted eat cookies" do
>     # ...
>   end
> 
>   # More consise!
>   it "permits admins to eat cookies" do
>     # ...
>   end
> end
> ```

styleguide/testing.md
```diff
@@ -110,8 +110,101 @@ it "is invalid without a description" do
...
+
+### Use `shared_contexts` and `shared_examples` to keep code DRY
+
+Shared examples are tests that test common behavior and are written in such a way that they can be used in several settings. A good example is our `"lesson_content"` example that is used to test the behavior of several models that are included as content in lessons.
+
+Shared contexts are common setup code that can be used to prepare tests in different settings. A common `shared_context` we use is the `logged in as a #{role}` context, which autocreates a `user` object of the desired role and a `company` that the user is associated with. Shared contexts are defined in `spec/support/shared_contexts.rb`. [Note: this specific example is defined with Capybara and is only usable with feature specs.]
+
+### RSpec Syntax
+
+- Use `specify` instead of `it` when it improves readability. This is common in our jobs specs, where `context` is used to refer to subclasses, and our model specs, where we test validations.
```
stevegrossi: > in our jobs specs, where `context` is used to refer to subclasses

I'm curious about this, as it contradicts something I've long assumed, which is that `describe` is for "things", and `context` is for "situations", e.g. `describe "a controller action"` but `context "when I'm logged in"`. With subclasses which are also "things", I would use nested `describe` blocks here:

```ruby
describe DoStuffJob do

  describe DoStuffJob::Subclass do
```

`DoStuffJob` and `DoStuffJob::Subclass` are both classes, so to use `describe` for one and `context` for the other seems inconsistent to me, suggesting a difference where there is none. What's your intuition here?

styleguide/testing.md
rreinhardt9: I think there is a potential to introduce a subtle flakiness here.

Let's say for example, you have something that puts the text "working" on the page until it receive a response from the server. We want that to be gone before we move onto the next step.
If we write some thing like `expect(page.has_no_content?("working")).to eq(true)` there is the possibility that because of how the timing works out, `page` at the point in time we check it here might still have "working" on it and so would fail (no waiting). If we use `expect(page).not_to have_content("working")` and it happens that we catch the page before "working" disappears, it will wait until that becomes true or until we hit the capybara timeout. In that second case, we avoid a flake where maybe the test ran really fast and "working" was still on the page. Additionally, if we run `expect(page).not_to have_content("working")` and it isn't on the page, it would pass right away; so I don't think there would be a time penalty in that case.

There is a link here that has some discussion around this https://robots.thoughtbot.com/write-reliable-asynchronous-integration-tests-with-capybara#looking-for-matching-css

I may be missing the point you were making though; is there a scenario where the `expect().to eq(true)` made more sense?

styleguide/testing.md
haleymnatz: Yes! Good call.

styleguide/testing.md
haleymnatz: Yes, especially because thoughtbot does that in the post I'm quoting from! Nice catch.

styleguide/testing.md
haleymnatz: The credit goes to @unixmonkey for this one! 👏

styleguide/testing.md
haleymnatz: Yes we should! Done. Thanks for catching the typo!

styleguide/testing.md
haleymnatz: My intuition is that I agree with you 100%. I was asked by someone else to add in the portion about `specify` and was referred to our Jobs as examples. Since I saw examples like the one in the code here so frequently, I just thought it was a Lessonly Job spec oddity. Perhaps those specs are just older and testing conventions were a little different when they were written?

If you'd like, I can abandon this approach and just follow the one like your comment around line 206.

styleguide/testing.md
haleymnatz: I share the same feelings as you here. Do you think there is ever a situation that warrants `specify`?

styleguide/testing.md
haleymnatz: I'd love that guidance too! 😆 What I was going for here was something that encompassed both services external to our code (like things that Webmock might be used for), but also other classes/methods that are called in our code.

For example, look at this code from `path_content_statuses_tree_query_decorator.rb`:

```
    def remaining_minutes_to_complete
      estimated_times = unfinished_lessons.uniq.map(&:estimated_minutes_to_complete)

      estimated_times.include?(nil) ? nil : estimated_times.reduce(:+)
end
```

In the specs for this, I stub the call to `unfinished_lessons` to reduce dependencies and improve testing in isolation. 

As for the question of stubbing external services - I don't quite feel qualified to give great advice on it, because it's not something I've had to do much of. I can look for articles to link to, but if you have any advice or thoughts on this, I'd love to consider adding it.

styleguide/testing.md
haleymnatz: @stevegrossi what are your thoughts here?

haleymnatz:
@stevegrossi Thanks for the thorough and thoughtful review! I've made the direct changes you requested, and left some questions/comments for further discussion on some of your thoughts.

styleguide/testing.md
stevegrossi: I personally don't find value in it, but I recognize that reflects my own experience and I'm interested in how others use and find value from `specify`.

styleguide/testing.md
stevegrossi: > Perhaps those specs are just older and testing conventions were a little different when they were written?

Or perhaps the conventions didn't exist!

> If you'd like, I can abandon this approach and just follow the one like your comment around line 206.

I never want my word to be the final say, but since this is your PR and you "100% agree", I'd of course be happy if you took that approach.

styleguide/testing.md
unixmonkey: FWIW, I never really use `specify` myself, nor have I seen it used much on projects I've been on, however, I don't have a problem with either form, so I don't see a reason to even call out preferring one over the other in the guide, unless we really don't like it and want to banish it from the codebase.

haleymnatz:
@stevegrossi I realize that some of the things I've added are more so "Best testing practices" rather than rules that we specifically adhere to. Yesterday in the CI Improvements retro, an idea came up to have a specific section devoted to that.

Do you think it's fine to leave those more "suggestion"- like points here, or should I make a separate page in the style guide for that?

styleguide/testing.md
stevegrossi: > I may be missing the point you were making though; is there a scenario where the expect().to eq(true) made more sense?

I think I'm in the same boat as @rreinhardt9 here. That Thoughtbot link is helpful and makes a persuasive case for preferring `expect(page).to have_content` over `expect(page.has_content?).to be true` because the first option "will pass immediately" and avoid waiting the full max_wait_time. @caryssaperez it sounds like you may be suggesting the opposite, but for the same reason. Is there an example where `expect(page.has_no_content?).to eq(true)` would be preferable?

PR #99, opened 2018-10-23
Add documentation for PrivateAttr, and clockwork

gardengeek:
Resolves [ch17240](https://app.clubhouse.io/lessonly/story/17240)
Based on [this](https://lessonly.slack.com/archives/G8Q5B0EVA/p1533828056000603) slack conversation, added a section in ruby.md around private accessors.

Also added a section in the Jobs documentation around how to schedule the background job.

### DIFF
--- a/styleguide/ruby.md
+++ b/styleguide/ruby.md
@@ -189,3 +189,51 @@ The latter introduces load order dependencies (`MyModule` must already be loaded
 ## Gemfile
 
 We try to use [pessimistic version constraints](http://guides.rubygems.org/patterns/#pessimistic-version-constraint) for all gems in the Gemfile. These use the `~>` operator followed by at least a major and minor gem version (e.g. `"~> 4.2"`). This prevents a gem from being accidentally upgraded by more than a minor version, and Gems following [Semantic Versioning](http://semver.org/) promise to introduce non-backwards-compatible changes only in major versions.
+
+## PrivateAttr vs private attr_reader vs instance variables
+
+Currently in the code there are three styles of accessors:
+ - Use of `PrivateAttr`
+ - Use of `private` with accessors below
+ - Just using instance variables for information hiding
+
+Some of the differences are described below. For declaring private accessors, both the PrivateAttr and having the `attr_reader` below `private` are acceptable options. Using instance variables should be avoided.
+
+### PrivateAttr
+
+The PrivateAttr functionality was added via [PR #2787](https://github.com/lessonly/lessonly/pull/2787) as a way to limit the public interface and use private methods rather than instance variables. One of the advantages of using PrivateAttr is that all of the instance variables will be listed at the top of the file for clarity.
+```
+class MyClass
+  extend PrivateAttr
+
+  private_attr_reader :foo, :bar, :baz
+
+  initialize(foo, bar, baz) do
+    @foo = foo
+    @bar = bar
+    @baz = baz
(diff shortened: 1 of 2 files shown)


styleguide/ruby.md
```diff
@@ -189,3 +189,51 @@ The latter introduces load order dependencies (`MyModule` must already be loaded
...
 We try to use [pessimistic version constraints](http://guides.rubygems.org/patterns/#pessimistic-version-constraint) for all gems in the Gemfile. These use the `~>` operator followed by at least a major and minor gem version (e.g. `"~> 4.2"`). This prevents a gem from being accidentally upgraded by more than a minor version, and Gems following [Semantic Versioning](http://semver.org/) promise to introduce non-backwards-compatible changes only in major versions.
+
+## PrivateAttr vs private attr_reader vs instance variables
+
+Currently in the code there are three styles of accessors:
+ - Use of `PrivateAttr`
+ - Use of `private` with accessors below
+ - Just using instance variables for information hiding
+
+Some of the differences are described below. For declaring private accessors, both the PrivateAttr and having the `attr_reader` below `private` are acceptable options. Using instance variables should be avoided.
```
case-eee: @gardengeek Thanks so much for writing this up 🎉 

> For declaring private accessors, both the PrivateAttr and having the `attr_reader` below `private` are acceptable options

For clarity sake, do you think we should decide which one of these we'd prefer folks on the team to use moving forward?

case-eee: I don't actually have a strong preference. Simply for the fact of having everything defined in one place within the file (at the top), I prefer using `PrivateAttr` so I'd probably suggest we stick with that one moving forward. I'm all for consistency 🎉

styleguide/ruby/job.md
```diff
@@ -68,3 +68,16 @@ end
 ```
 
 In this case, if `OldJob` is already queued, it will still run as it extends `NewJob`.
+
+### Scheduling a job
+
+Lessonly uses clockwork to schedule background jobs. The job will need to be added to `lib/clock.rb`.
+
```
haleymnatz: I've never written a job or used clockwork before, so for someone like me, it could be helpful to have a link included to the documentation? I see that there are several forks of repos called "clockwork", so I'm not sure which one we're using :/

(As a sidenote, it would be great if the Gemfile listed gems by the full repo name - like `thoughtbot/factory_bot` instead of just `factory_bot`!)

styleguide/ruby.md
gardengeek: I have updated the styleguide to say that PrivateAttr is the preferred method. Want to take another look and see if my changes look OK?

styleguide/ruby/job.md
gardengeek: I have added a couple of links that might be helpful. I'm not actually sure which fork we are using either but figure the link to the original can be used to find the others.

gardengeek:
@case-eee I have updated the links to point to the version of clockwork that we are currently running, and the page has links to more current versions when we upgrade the gem. Want to take another look?

styleguide/ruby/job.md
```diff
@@ -68,3 +68,21 @@ end
...
+To kick-off the background job from the command line:
+```
+bundle exec clockwork lib/clock.rb
+```
+
+Useful links:
```
case-eee: 👍

PR #105, opened 2018-12-07
Adding new page style to lessonly.github.io

wuz:
Hey all! I took some of the time I had before the Holiday party to put together a new style/design for the engineering page we have. I'd love to see us use this more, and I thought a makeover might help with that 💅. I've got the mobile site working a bit better and I added our open source projects as well!

Thanks!


armilam:
This is looking great! Thank you for taking the initiative! I might recommend getting feedback from Marketing on this one, since it's turning into something that displays our brand. Though, I can see an argument for doing that as a second iteration.

I did find this. I'd recommend removing the underlines on the nav bar links, since www.lessonly.com does not underline those.

wuz:
Definitely! I really wanted to discuss with Marketing, but I also wanted to get _something_ out and have some momentum here. 

> I'd recommend removing the underlines on the nav bar links, since www.lessonly.com does not underline those.

Sounds good!

wuz:
@stevegrossi I think that would be a great couple of first steps! I thought I had matched the colors, but looks like I didn't. I submitted this to marketing for review, but haven't heard back yet, unfortunately. Getting the fonts and colors is a great step forward.

stevegrossi:
Given the current site bears very little relation to our brand, I'll make the call that we don't need to wait on marketing's feedback for something like this (though we should consider it if they do respond), so long as we're confident that it's on-brand. With the same font and colors, I'd have that confidence.

wuz:
Alrighty! We've got the fonts hooked up and I added a header to the index page to fit with our *Do Better Work* company wide push. What do you think?

screencapture-localhost-4000-2019-02-27-11_59_28

stevegrossi:
Hot dang that screenshot looks good! I'll take a closer look this week, but 😗👌✨

armilam:
Wow, that is looking goooooooood

>

wuz:
@stevegrossi just wanted to bump this 😄 No rush, just don't want it to get lost!

stevegrossi:
It's on my list and won't get lost! I apologize for letting it linger, it's been a week of surprises.

wuz:
> It's on my list and won't get lost! I apologize for letting it linger, it's been a week of surprises.

No problem at all! Like I said, no rush. Thanks for looking at it 😄

_config.yml
```diff
@@ -2,11 +2,17 @@
...
+  Product and engineering at Lessonly
 baseurl: "" # the subpath of your site, e.g. /blog/
 url: "http://lessonly.github.io" # the base hostname & protocol for your site
 twitter_username: lessonly
-github_username:  lessonly
+github_username: lessonly
+collections:
+  open_source:
+    output: false
+  styleguide:
```
stevegrossi: I like this decision: it'll help us keep style guide pages separate from other random ones 👍

PR #110, opened 2019-02-20
📚 Add ERB view rule to Rails styleguide

### LINKED ISSUE #104: Rails: prefer JSON endpoints
I would like to propose that we start moving towards a more [JAMstack](https://jamstack.org/)/[Immutable Web App](https://immutablewebapps.org/) style of building the application. With that in mind, I think establishing a rule around whether or not we pass data through ERB or load data via an JSON endpoint. Basically, instead of:

```ruby
<%= react_component("TestComponent", props: {
    data: SomeSerializer.new(@data).as_json,
  }) %>
```

```js
import React, { Component } from 'react';

class TestComponent extends Component {
	render() {
		const { data } = this.props;
		return <Something data={data} />
	}
}

```

we would just render the React component and in that component we would do a fetch/request:

```ruby
<%= react_component("TestComponent") %>
```

```js
import React, { Component } from 'react';

class TestComponent extends Component {
	constructor(props) {
		super(props);
		this.state = { 
			loading: true,
			data: {},
		}
	}
	componentDidMount() {
		request('/someserialized/data.json').then((data) => {
			this.setState({ data, loading: false });
		});
	}
	render() {
		const { data } = this.state;
		return <Something data={data}/>
	}
}

```

Arguments in favor:

- This would allow us at some point to split the React rendering into a separate SPA, which could be served from S3 and heavily cached.
- This allows Rails to handle just rendering data, without having to deal with ERB's or marked up views.
- This is very much the "new style" of building applications (which could be good or bad, but I would argue is good)

I'd love to hear others thoughts!

wuz:
Fixes #104

### DIFF
--- a/styleguide/rails.md
+++ b/styleguide/rails.md
@@ -58,13 +58,67 @@ User.where(id: lesson.progresses.select(:user_id))
 emails = company.users.pluck(:name, :email)
 ```
 
-Of course, if your *end goal* is an array of simple attributes, that's exactly what `pluck` is for.
+Of course, if your _end goal_ is an array of simple attributes, that's exactly what `pluck` is for.
 
 ### Views
 
 #### Avoid referencing instance variables in partials.
 
 It's alright for `show.html.erb` to reference the `@lesson` declared in `LessonsController#show`, but if it calls out to `render "stats"`, `_stats.html.erb` should not know about `@lesson`. Instead pass it in directly: `render "stats", lesson: @lesson`. Eventually, we'll want to use that partial in another context where `@lesson` isn't defined: being explicit now saves us time later.
+
+#### Avoid ERB views and server-rendered markup when possible.
+
+Instead, we prefer to mount React components that fetch the data they need to render themselves via JSON endpoints.
+
+Example:
+
+Instead of:
+
+```ruby
+<%= react_component("TestComponent", props: {
+    data: SomeSerializer.new(@data).as_json,
+  }) %>
+```
+
+```js
+import React, { Component } from "react";
+
+class TestComponent extends Component {
+  render() {
+    const { data } = this.props;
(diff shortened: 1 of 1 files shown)


rreinhardt9:
I love this update; it helps push us towards more reusable self contained components and better user experiences because we can deal better with loading states, reduce time to first paint, and such ❤️

styleguide/rails.md
```diff
@@ -58,14 +58,68 @@ User.where(id: lesson.progresses.select(:user_id))
...
 
 ### Views
 
 #### Avoid referencing instance variables in partials.
 
 It's alright for `show.html.erb` to reference the `@lesson` declared in `LessonsController#show`, but if it calls out to `render "stats"`, `_stats.html.erb` should not know about `@lesson`. Instead pass it in directly: `render "stats", lesson: @lesson`. Eventually, we'll want to use that partial in another context where `@lesson` isn't defined: being explicit now saves us time later.
 
+#### Avoid ERB views and server-rendered markup when possible.
+
+Instead, we prefer to mount React components that fetch the data they need to render themselves via JSON endpoints.
```
stevegrossi: This is a fabulous addition ❤️ Should we say a few words as to why this is preferred so future readers get the context? e.g.

> This approach reduces [time to first paint](https://developers.google.com/web/tools/lighthouse/audits/first-meaningful-paint), allows pages to be more resilient to errors, and works toward our end goal of separating data (server-side) and presentation (client-side).

(That last bit is open to discussion, but it's a goal of mine at least.)

wuz: Sure! Do we have a standard for entries in the style guide? Might be something to work on in the future!

wuz:
Added the reasoning that Steve recommended!

PR #114, opened 2019-03-04
Add entry on i18n constants

joshmfrankel:
# Why

We learned that using frozen constants and translations doesn't allow for proper loading and multilingual configurations. https://github.com/lessonly/lessonly/pull/6335. We need to evaluate translations everytime in case a client switches languages.


styleguide/rails.md
```diff
@@ -3,7 +3,29 @@ layout: page
...
+# Avoid
+class SomeClass
+  UNAUTHORIZED_JSON_RESPONSE = { error: I18n.t("common.unauthorized") }.freeze
+end
+
+# Recommended
+class SomeClass
+  def unauthorized_json_response
+    { error: I18n.t("common.unauthorized") }
+  end
```
unixmonkey: I think this example could be simplified further to make it clearer that it doesn't have anything to do with the JSON, the hash, the class or the freeze. Here's my take:

```ruby
# Avoid - This is evaluated once at load time and will always be in the default language (English).
ERROR_MESSAGE = I18n.t("common.error")

# Recommended - This is evaluated every time it's called with the locale of the current request.
def error_message
  I18n.t("common.error")
end
```

With these tweaks, I think it might be clear enough to remove the paragraph, too. I'm not sure whether this should be in the Rails guide, or the Ruby one, and whether it should be the first thing in the file.

FYI, my plan is to create an entire guide just for translation-related stuff. Maybe we should split this off into a new file as the first entry? We also have the same problem in JavaScript-land, and I think it might be beneficial to explain them both at the same time, rather than splitting it into Ruby/Rails/JavaScript.

```js
// Avoid - This is hardcoded at build time, before translations are even loaded.
const ERROR_MESSAGE = t('common.error');

// Recommended - This is evaluated every time with the locale of the current request.
const errorMessage = () => t('common.error');

// or with a getter function
const MESSAGES = {
  get ERROR() {
    return t('common.error');
  }
};
```

What do you think?
Thank you for taking the initiative to document this gotcha while it's still fresh for you. 🥇

joshmfrankel: > FYI, my plan is to create an entire guide just for translation-related stuff. Maybe we should split this off into a new file as the first entry? We also have the same problem in JavaScript-land, and I think it might be beneficial to explain them both at the same time, rather than splitting it into Ruby/Rails/JavaScript.

Nice! I love this idea. I'd be up for making a new internationalization entry. I also agree with simplifying the examples above :100:

joshmfrankel: Added :+1:

styleguide/rails.md
```diff
@@ -3,7 +3,7 @@ layout: page
 title: Rails Style Guide
 ---
 
-While we don’t follow it explicitly, the [community Rails Style Guide](https://github.com/bbatsov/rails-style-guide) (also edited by Bozhidar Batsov) is a solid refrence when in doubt.
+While we don’t follow it explicitly, the [community Rails Style Guide](https://github.com/bbatsov/rails-style-guide) (also edited by Bozhidar Batsov) is a solid reference when in doubt.
```
joshmfrankel: Just a spellling change

joshmfrankel:
This one is ready for merging :rocket:

PR #115, opened 2019-03-12
✍ DOC: Add entry for explicit enum mappings

joshmfrankel:
# Proposed change

Instead of non-specified mappings:

`enum scenario_type: %i(webcam_recording screen_recording audio_recording`

Prefer to use explicit mapping

```
enum scenario_type: {
      webcam_recording: 0,
      screen_recording: 1,
      audio_recording: 2,
```

This avoids re-ordering or removal of existing enum keys which leads to inconsistent data returned.

#  Context
If the order of enum types changes then it maps to a different integer. Meaning changing model code could essentially change what type is returned based on the database column. 

For example, if someone were to switch Answer's question_type order slightly:

``` ruby
# Current implementation
enum question_type: %i(multiple_choice free_response multiple_select survey webcam_recording screen_recording written_response email_composition email_reply chat ticket_handling knowledge_check audio_recording)

development > Answer.question_types
=> {"free_response"=>0,
 "multiple_choice"=>1,
 "multiple_select"=>2,
 "survey"=>3,
 "webcam_recording"=>4,
 "screen_recording"=>5,
 "written_response"=>6,
 "email_composition"=>7,
 "email_reply"=>8,
 "chat"=>9,
 "ticket_handling"=>10,
 "knowledge_check"=>11,
 "audio_recording"=>12}

development > Answer.find(61742)
=> #<Answer:0x0000558e5b64e530
 id: 61742,
 progress_id: 36394,
 question_id: 313,
 question_type: "audio_recording",
 question_text: "Audio",

# Switch the order of audio_recording and free_response
enum question_type: %i(audio_recording multiple_choice multiple_select survey webcam_recording screen_recording written_response email_composition email_reply chat ticket_handling knowledge_check free_response)

development > reload!
development > Answer.question_types
=> {"audio_recording"=>0,
 "multiple_choice"=>1,
 "multiple_select"=>2,
 "survey"=>3,
 "webcam_recording"=>4,
 "screen_recording"=>5,
 "written_response"=>6,
 "email_composition"=>7,
 "email_reply"=>8,
 "chat"=>9,
 "ticket_handling"=>10,
 "knowledge_check"=>11,
 "free_response"=>12}

development > Answer.find(61742)
=> #<Answer:0x0000558e5e3286f0
 id: 61742,
 progress_id: 36394,
 question_id: 313,
 question_type: "free_response", # Now it says the Answer is a free_response
 question_text: "Audio",
```

From the above free_response was originally mapped to integer 0 but after changing its location and audio_recording is now mapped to integer 12. Looking up records before and after highlights the problem.

By explicitly setting the mapping even changing the order of types wouldn't effect the returned data. Only by changing the actual integer associated with the key would impact results. Basically, a simple way to ensure a little better data integrity until we decide to use PostgreSQL native enum types :sunrise: 

Example explicit mapping

``` ruby
    enum scenario_type: {
      webcam_recording: 0,
      screen_recording: 1,
      audio_recording: 2,
      written_response: 3,
      email_composition: 4,
      email_reply: 5,
      chat: 6,
      ticket_handling: 7
    }
```

# Rails Documentation

> Note that when an array is used, the implicit mapping from the values to database integers is derived from the order the values appear in the array. In the example, :active is mapped to 0 as it's the first element, and :archived is mapped to 1. In general, the i-th element is mapped to i-1 in the database.

> Therefore, once a value is added to the enum array, its position in the array must be maintained, and new values should only be added to the end of the array. To remove unused values, the explicit hash syntax should be used.

https://api.rubyonrails.org/v5.2.2/classes/ActiveRecord/Enum.html


stevegrossi:
Great suggestions for taking this from 👍 to 💯, @unixmonkey!

joshmfrankel:
Cleaned up the examples and added the documentation link. Also merged in master :rocket:

stevegrossi:
Whoops, could swear I clicked "Merge" on this. Onward!

PR #117, opened 2019-03-26
Remove irrelevant javascript translation rules

wsmd:
This section is no longer relevant. This was resolved in [ch24334].

You should now be able to use the javascript `t()` however you want. Even outside of a React render method or before the page is fully loaded.

### DIFF
--- a/_styleguide/translations.md
+++ b/_styleguide/translations.md
@@ -8,8 +8,6 @@ main: true
 
 #### Avoid using I18n within constants
 
-**Ruby context**
-
 ```ruby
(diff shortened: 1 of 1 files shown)


unixmonkey:
@wsmd I don't think we need the `**Ruby context**` line any longer, either.

wsmd:
@unixmonkey do you mean just the one line? or the entire rule?

unixmonkey:
@wsmd Just the one line and the blank line below it. No need for separate section delimiters, when there's only one.

PR #121, opened 2019-04-16
Add another option for active record model validation messages

armilam:
This file already contains great information about preloading translations directly into English. I stumbled into a similar pattern we're using which also preloads the English translation for error messages.

This PR adds another "Avoid" example and a "Recommended" example to go along with it.

### DIFF
--- a/_styleguide/translations.md
+++ b/_styleguide/translations.md
@@ -19,3 +19,17 @@ def error_message
   I18n.t("common.error")
 end
 ```
+
+#### Avoid using I18n outside of a method or lambda
+
+```rb
+# Avoid
+# This is evaluated once at load time and will always be in the default language (English).
+validates :name, presence: { message: I18n.t("common.error") }
+
+# Recommended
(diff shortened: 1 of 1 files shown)


unixmonkey:
Another good way to avoid using a proc or lambda here, is to use the built-in ActiveRecord framework error message key lookup paths:

https://guides.rubyonrails.org/i18n.html#error-message-scopes

By namespacing your error messages under special key lookup locations like `en.errors.messages` like this:

```yaml
en:
  errors:
    messages:
      custom_message: "This a custom message"
      oh_no: "cannot be saved with %{value} for %{attribute} of %{model}"
```

```ruby
class Thing < ApplicationModel
  validates :name, presence: { message: :oh_no }
  validate :looks_good_to_me

  private

  def looks_good_to_me
    errors.add(:base, :custom_message)
  end
end

thing = Thing.new
thing.valid? # => false
thing.errors.full_messages # => ["Name cannot be saved with  for Name of Thing", "This is a custom message"]
```

Issue #58, opened 2017-01-18
API principles


stevegrossi:
Our API has grown organically, which is cool but we've learned some things which would probably be good to make conventions. For instance, we learned not to just tack additional, expensive data (e.g. completion stats) on to otherwise fast, frequently-accessed endpoints (like `/courses/:id`). What else have we learned that we'd like to record for our future selves?

## Ideas

- http://www.vinaysahni.com/best-practices-for-a-pragmatic-restful-api
- https://github.com/interagent/http-api-design/blob/master/en/SUMMARY.md

stevegrossi:
Oh, and of course, *always* return a JSON object, never an array, to allow for future metadata without breaking backward compatibility.

For example, if an endpoint returns `[...]` and we later decide to paginate it, we'll need to return an object instead of an array like `{ page: 1, totalPages: 5, lessons: [...]}` and any clients expecting an array will 💥 . We can avoid this by always passing an object `{ lessons: [...] }` even if the only key points to an array. It's just a way of not reducing our options unnecessarily in the future.

@lessonly/developers I'd love some more thoughts here.

wsmd:
> never an array, to allow for future metadata without breaking backward compatibility.

Totally agree :+1:

We did something similar with a `POST` request computing the unique users from a selection.

**Request body**
```js
{
  assignees: [...],
}
```

**Response body**
```
{
  count: ...
}

joshmfrankel:
> always return a JSON object

Agreed.

armilam:
Perhaps tangentially related, but we should always be updating the API docs along with API updates

stevegrossi:
We've started sending JSON for POST/PUT requests (rather than form data). Presumably we did this for a reason. Should we always be doing this? I think so.

stevegrossi:
This is Ruby-related, but our new `V1_1` api has a new pattern for serialization (they live in the API, not in the model) that I think is an improvement. See more: https://github.com/lessonly/lessonly/pull/3676/files#r115732152

Perhaps we should document this, too.

PR #122, opened 2019-04-19
Improve json api styleguide

joshmfrankel:
Just a slight improvement to this section with an example highlighting what to do / not to do

### DIFF
--- a/_styleguide/json.md
+++ b/_styleguide/json.md
@@ -14,6 +14,8 @@ Rails controllers handle JSON `params` as easily as form data. JSON is a more co
 
 Working with JSON objects with named keys is easier than working with unnamed Arrays, Integers, and Strings. It is also more understandable for future development.
 
+For example, if an endpoint returns `[...]` and we later decide to paginate it, we'll need to return an object instead of an array like `{ page: 1, totalPages: 5, lessons: [...]}` and any clients expecting an array will 💥 . We can avoid this by always passing an object `{ lessons: [...] }` even if the only key points to an array.
+
 ```js
 // Avoid returning arrays
 [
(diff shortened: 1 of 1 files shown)


_styleguide/json.md
```diff
@@ -6,6 +6,30 @@ main: true
...
+  attempts: [
+    {
+      id: 1
+    },
+    {
+      id: 2
+    }
+  ]
+}
+```
```
stevegrossi: I'm grateful for this addition, but I think the rule you're replacing is about something quite different, and would suggest you make your change as an addition rather than a replacement. It can certainly be clearer, but the guideline you're modifying, "Use JSON for request bodies." is specifically about the data we send _to_ the server in requests, and is more about the type of data, e.g. to prefer submitting JSON bodies (as `application/json`) rather than HTML form data (as `application/x-www-form-urlencoded`), whereas the examples here (and the addition I think you're trying to make) are about the shape of JSON we return _from_ the server.

Also, the examples, which contrast "Avoid returning arrays" with "Recommend returning JSON", may be confusing because arrays _are_ valid JSON ([so are bare strings and numbers](https://developer.mozilla.org/en-US/docs/Learn/JavaScript/Objects/JSON#Other_notes), I just learned!). From your examples, I think the specific guideline would be to always return JSON objects, rather than JSON arrays. Does that seem right?

joshmfrankel: Ah, good catch. I had a nagging thought that I was not understanding the current style correctly and should have asked for clarification first. 

I agree with your suggestions above :+1: I'll implement those and thanks for the feedback

joshmfrankel: Alright, I've made the changes and included a bit about what you said regarding Strings and Integers :tipping_hand_man:

_styleguide/json.md
```diff
@@ -6,6 +6,41 @@ main: true
 
 Unless otherwise noted, these rules apply only to internal JSON endpoints, not necessarily [our public API](https://docs.lessonly.com).
 
-## Use JSON for request bodies.
+## Use JSON for request bodies
 
-Rails controllers handle JSON `params` as easily as form data, and JSON is more concise and flexible (especially when handling arrays and objects). Also, our endpoints always return JSON, so it’s more consistent also always to send it.
+Rails controllers handle JSON `params` as easily as form data. JSON is a more concise and flexible format for our endpoints to remain consistent. Because of this we prefer to send Content-Type `application/json` in requests.
```
joshmfrankel: Slight editing above to improve clarity

_styleguide/json.md
stevegrossi: > Working with JSON objects with named keys is easier than working with unnamed Arrays, Integers, and Strings. It is also more understandable for future development.

In my experience, this is true, but for the sake of future colleagues I think it's important to explain _why_ things are easier or more understandable. Why do you find objects easier and more understandable in these examples?

I'm not suggesting you copy it, but when I originally logged this issue in #58, here's [why I proposed this rule](https://github.com/lessonly/lessonly.github.io/issues/58#issuecomment-273529176):

> ...to allow for future metadata without breaking backward compatibility.
> 
> For example, if an endpoint returns `[...]` and we later decide to paginate it, we'll need to return an object instead of an array like `{ page: 1, totalPages: 5, lessons: [...]}` and any clients expecting an array will 💥 . We can avoid this by always passing an object `{ lessons: [...] }` even if the only key points to an array. It's just a way of not reducing our options unnecessarily in the future.

joshmfrankel:
@stevegrossi I went ahead and copied your explanation into this styleguide issue. I also added a second section dealing with integer keys in json being automatically sorted by browsers and to avoid using them :)

PR #129, opened 2019-10-22
Prefer being explicit about controller behavior

### LINKED ISSUE #63: Rails: avoid `except` in callbacks
e.g. `before_action :do_something_expensive, except: :index`, because it silently adds new behavior in places where it may not be needed. We should always use `only` so we only get behavior where we actually need it.

For things we do 90% of the time (require login) but don't want to do in rare cases, we should prefer a separate, explicit `skip_before_action` over `before_action except:`

stevegrossi:
# Why

Resolves #63

# What

Screen Shot 2019-10-22 at 6 47 41 AM

# Alternatives Considered

An enforceable Rubocop rule is always better than a Style Guide recommendation, but [`rubocop-rails` does not have one for this](https://github.com/rubocop-hq/rubocop-rails/tree/master/lib/rubocop/cop/rails). I'd love to learn how to write custom cops, but didn't want to let that hold up this valuable addition to the Style Guide.


_styleguide/rails.md
```diff
@@ -143,6 +143,27 @@ class TestComponent extends Component {
...
+# Better: only the action that needs this behavior will get it
+before_action :find_the_thing, only: :show
+
+# Even better: only the actoon that needs this behavior will get it,
+# AND it’s obvious where the behavior is coming from
+def show
+  find_the_thing
+  # ...
+end
+```
```
stevegrossi: Ultimately, [I'm with Thoughtbot](https://thoughtbot.com/blog/before-filter-wisdom) that `before_action`s and their ilk are all cost (in terms of misdirection) and no benefit. They make it necessary to look in several places in a file to determine what's happening in a single controller action, impeding understandability and productivity and making bugs more likely. We'd benefit from a future in which we no longer use them.

But we make extensive use of them today and I didn't want to introduce the cognitive dissonance of outlawing them here, so I've started with recommending not using them as an alternative, not a mandate.

rreinhardt9:
I don't really feel strongly one way or another and agree with that `_action` callbacks can lead to indirection. I'm curious though, what about times where you do want to enforcing something that you have to opt out of? For example... pundit will enforce that you authorize every action and if you need to do something different you have to explicitly opt out with something like `after_action :verify_authorized, except: [:get_token]`? It seems like there might be cases where it could be helpful then?

Part of the hazard with it being surprising could come as non restful routes are added to controllers as well. It's a little easier to keep track of which one's it's applying to if your controller only handles restful routes... but if the controller starts getting a bunch of additional non restful routes I could see where that would start to obfuscate what is going on (and maybe be a sign that a new controller should be created for those routes 😬)

_styleguide/rails.md
```diff
@@ -143,6 +143,27 @@ class TestComponent extends Component {
...
+Applying an `_action` callback to all controller actions `except` some will automatically (and perhaps unexpectedly) opt us into this behavior in all future controller actions. This violates the [Principle of Least Astonishment](https://en.wikipedia.org/wiki/Principle_of_least_astonishment) and has led to bugs in the past.
+
+```ruby
+# Not good: actions added later will make this query even when unnecessary
+before_action :find_the_thing, except: :index
+
+# Better: only the action that needs this behavior will get it
+before_action :find_the_thing, only: :show
+
+# Even better: only the actoon that needs this behavior will get it,
```
armilam:
```suggestion
# Even better: only the action that needs this behavior will get it,
```

unixmonkey:
Please forgive me for dropping in on this slightly unrelated topic, but I was reminded by this of an even bigger gotcha with `*_action` controller methods I've run into a few times. If you combine conditions, it can have very surprising effects.

If you write up a filter like this:
```ruby
before_action :require_authorization,
  except: :show,
  if: -> { current_user.present? }
```

In addition to it being a bit complicated for a filter, it won't do what it looks like, it will ignore the `:except`, and run the `if`, even on the show action.

And when using `:only` with a condition:
```ruby
before_action :render_404,
  only: :show,
  if: -> { current_user.nil? }
```

Here, the `if` gets ignored, so a nil user could still see the `show` action.

This is because `:only` and `:accept` options get converted into either an `if` or `unless` Proc, overriding the one you specified.

This [is documented here](https://api.rubyonrails.org/classes/AbstractController/Callbacks/ClassMethods.html#method-i-_normalize_callback_options), but violates the principle of least astonishment IMO. I'd rather it raise an error if you end up accidentally doing this.

stevegrossi:
> For example... pundit will enforce that you authorize every action and if you need to do something different you have to explicitly opt out with something like after_action :verify_authorized, except: [:get_token]? It seems like there might be cases where it could be helpful then?

I like that Pundit provides a way to explicitly opt out of this in a particular action. I would find this:

```ruby
def get_token
  skip_authorization
  # ...
end
```

far preferable to `except: [:get_token]` at the top of the file, wouldn't you?

> Part of the hazard with it being surprising could come as non restful routes are added to controllers as well.

Right there with ya, @rreinhardt9  😄! https://github.com/lessonly/lessonly.github.io/issues/98

stevegrossi:
Thanks @unixmonkey! I wasn't aware of that gotcha, but it's a doozie 😬 

And @armilam, accepting your spelling-error-fix dismissed your approval, so please re-approve (no rush) when you get a chance.

_styleguide/rails.md
joshmfrankel: I'm agreed that before_actions can be quite dangerous. That being said what I like to do is explicitly opt into them using `only: `. That way future actions don't have mysterious side-effects only the ones specified. Maybe we want to add something about this as an alternative in case someone is thinking of using a callback

joshmfrankel: Yup!

PR #130, opened 2019-11-01
Bump ruby version to 2.6.5

joshmfrankel:
2.4.2 => 2.6.5


.tool-versions
```diff
@@ -0,0 +1 @@
+ruby 2.6.5
```
stevegrossi: Thanks for adding this 👍

PR #134, opened 2020-03-19
Add a new Style Guide section on Database best practices

### LINKED ISSUE #88: DB styleguide
Source: https://lessonly.slack.com/archives/C69BRKDL7/p1515429653000357

Adding or removing large indexes risks failure due to `PG::TRDeadlockDetected` issues when there are constantly queries making use of that index. We can add `algorithm: :concurrently` to get around that, (which requires disabling the wrapper transaction with `disable_ddl_transaction!`), e.g.

```ruby
class AddCompoundIndexToAssignments < ActiveRecord::Migration
  disable_ddl_transaction!

  def change
    add_index :assignments, [:assignee_id, :assignable_id, :assignable_type], unique: true, algorithm: :concurrently, name: 'index_unique_assignee_id_assignable_id_and_assignable_type'
  end
end

```

stevegrossi:
# What

Our app depends on our database(s), and if we want them to treat us well, we need to treat them well too. We should start documenting what we've learned about how best to do that.

Resolves #88

### DIFF
--- a/_styleguide/database.md
+++ b/_styleguide/database.md
@@ -1,0 +1,27 @@
+---
+layout: styleguide
+title: Database Style Guide
+main: true
+---
+
+## Defer to the PostgreSQL wiki about what not to do.
+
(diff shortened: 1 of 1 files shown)


stevegrossi:
Thanks! With 2 approvals, I'll go ahead and merge this.

PR #135, opened 2020-05-12
Update json style guide for camel case guidelines

ccharris:
Per this slack conversation [here](https://lessonly.slack.com/archives/G8Q5B0EVA/p1588182971393900
) we agreed that we would only use camelCase for all JSON, so that React can use only camel case and then we would transform it into snake case in Ruby.

Relates to this [story](https://app.clubhouse.io/lessonly/story/45280/automatically-snake-case-camelcased-json-params) to make turning camel case into snake case more automatic.


_styleguide/json.md
```diff
@@ -6,6 +6,28 @@ main: true
 
 Unless otherwise noted, these rules apply only to internal JSON endpoints, not necessarily [our public API](https://docs.lessonly.com).
 
+## Use camel case 
+
+We should use camel case in JSON sent to or received from our API.
```
haleymnatz, line 11: My only suggestion which actually doesn't matter so you don't have to do it would be to write `camelCase` here :)

README.md
```diff
@@ -7,7 +7,7 @@ Powered by [Jekyll](https://github.com/jekyll/jekyll) and hosted by [Github Page
     $ git clone git@github.com:lessonly/lessonly.github.io.git
     $ cd lessonly.github.io
     $ bundle install
-    $ jekyll serve
+    $ bundle exec jekyll serve
```
stevegrossi, line 10: Good catch, thanks for making this better!

_styleguide/json.md
```diff
@@ -31,7 +53,7 @@ For example, if an endpoint returns `[...]` and we later decide to paginate it,
 5
 
 // Avoid returning strings
-"some_return_value"
+"someReturnValue"
```
stevegrossi, line 56: Yay consistency! 🎉

_styleguide/json.md
ccharris: Haha we actually originally had them all as `camelCase` and I undid some because I worried it was too much, I'll fix!

PR #139, opened 2020-08-18
Add note about HTTP status codes

case-eee:
This resolves [ch52097](https://app.clubhouse.io/lessonly/story/52097/add-note-about-http-status-codes-to-style-guide) and adds clarity around utilizing the most appropriate HTTP code when creating JSON endpoints!

### DIFF
--- a/_styleguide/json.md
+++ b/_styleguide/json.md
@@ -108,3 +108,12 @@ JSON objects that use integer keys which also depend on sorting are automaticall
   ]
 }
 ```
+
+## Use the most appropriate HTTP status code in your response
+
+Utilize successful `200 Ok` codes **only** to indiciate success. 
+
+Avoid returning a `200` with a body of `{success: false}` - this is confusing and requires our client-side code has to go out of its way to handle this mixed message from the server.
+
+
+
--- a/_styleguide/json.md
+++ b/_styleguide/json.md
@@ -111,12 +111,11 @@ JSON objects that use integer keys which also depend on sorting are automaticall
 
 ## Use the most appropriate HTTP status code in your response
 
-Utilize successful `200 OK` codes **only** to indiciate success. 
+Utilize successful `200 OK` codes **only** to indicate success with a proper body that matches `Content-Type` header. 
 
 Avoid returning a `200` with a body of `{success: false}` - this is confusing and requires our client-side code to go out of its way to handle this mixed message from the server.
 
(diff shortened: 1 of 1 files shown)


_styleguide/json.md
```diff
@@ -108,3 +108,15 @@ JSON objects that use integer keys which also depend on sorting are automaticall
   ]
 }
 ```
+
+## Use the most appropriate HTTP status code in your response
```
emmoistner, line 112: Yes! 🙌

_styleguide/json.md
```diff
@@ -108,3 +108,15 @@ JSON objects that use integer keys which also depend on sorting are automaticall
...
+
+Utilize successful `200 OK` codes **only** to indiciate success.
```
emmoistner, line 114: 🤔 I think we would add another layer here to cover the 204 no content issue we run into.  Also noticed a typo 

```suggestion
Utilize successful `200 OK` codes **only** to indicate success with a proper body that matches `Content-Type` header. 
```

_styleguide/json.md
```diff
@@ -108,3 +108,15 @@ JSON objects that use integer keys which also depend on sorting are automaticall
   ]
 }
 ```
+
+## Use the most appropriate HTTP status code in your response
+
+Utilize successful `200 OK` codes **only** to indiciate success. 
+
```
emmoistner, line 115: We could add another line discussing a little bit about suggestion above. If there is no proper body that matches the content-type (for example `application/json`) return `204 No Content` or if the response can't return body because it is asynchronously processing we could return a `202 Accepted`

rreinhardt9 (approved):
Seems sensible to me! That integer key and order thing has bit me in the past 😆 Using semantic HTTP codes also seems like a good general idea.

One potential distinction is, HTTP codes have to do with the status of the request... but you will have situations where the request was successful (200) but there is something you want to communicate back to the end user, like a validation issue ("that isn't a valid URL that you supplied" for example). For that, I don't think we'd want to return something like a 400 bad request because the request was just fine, it's just that our business logic requires something else that we want to communicate back to the user. That's a subtle distinction I see there.

By the same token if you supplied a value of `1234` (an integer) as a value to a parameter expecting a String, that would be a time to return 400 bad request because the request itself is not appropriate based on what we were expecting.

emmoistner (approved):
👍

case-eee:
Thanks for the review y'all!! I'm going to go ahead and merge this 🎉

PR #141, opened 2020-12-02
Update our recommendation around foreign-key constraints

stevegrossi:
Screen Shot 2020-12-02 at 10 56 57 AM

# What

We used to recommend always enforcing referential integrity in the database with `FOREIGN KEY` constraints, but ran into performance issues with the `users` table having 60+ of them and deleting users being unsustainably slow as a result. We now recommend not using FK constraints unless we have a compelling reason to.

# See Also

- Initial [post in Slack](https://lessonly.slack.com/archives/G8Q5B0EVA/p1594301004090400) about the moratorium
- Slack post [on this Style Guide/Lesson update](https://lessonly.slack.com/archives/C8LGVESDC/p1606858496380100)

### DIFF
--- a/_styleguide/database.md
+++ b/_styleguide/database.md
@@ -8,13 +8,13 @@ main: true
 
 The PostgreSQL Wiki has a handy [“Don’t Do This”](https://wiki.postgresql.org/wiki/Don't_Do_This) page. While there are exceptions when most of these discouraged practices may be safe or even necessary, we should avoid them unless we’re sure we’re the exception.
 
-## Enforce data integrity at the database level whenever possible.
+## Enforce data integrity with unique indexes and not-null constraints
 
 PostgreSQL gives us a range of tools to ensure invalid data isn’t just unlikely but impossible:
 
 - Use `UNIQUE` `CONSTRAINT`s/`INDEX`es (`unique: true` in migrations) to prohibit “duplicate” data. For instance, if it doesn’t make sense for a Lesson to be tagged with “Sales” _twice_, we want a unique composite index on the `tag` and `taggable`.
 - Use `NOT NULL` (`null: false` in migrations) to guarantee necessary relationships (e.g. make `company_id` non-nullable on tables that must be associated with a company) or attributes (if a `title` is required, make it `NOT NULL`).
-- Use `FOREIGN KEY` constraints to ensure that references to other tables like `company_id` point to records that actually exist, and specify `ON DELETE` `CASCADE` or `NULLIFY` behaviors when those records are removed. (Be mindful of performance when adding foreign key constraints: see [this migration](https://github.com/lessonly/lessonly/pull/5390/files#diff-bf0c0f95fbacafc133567b5e36e2f289R20) for an illustration the challenges and a way around them.)
+- **Update:** we no longer recommend adding `FOREIGN KEY` constraints by default, as we’ve run into performance issues on tables like `users` that have very many of them. However, we should still consider them in cases where referential integrity is essential. (And be mindful of performance when adding foreign key constraints: see [this migration](https://github.com/lessonly/lessonly/pull/5390/files#diff-bf0c0f95fbacafc133567b5e36e2f289R20) for an illustration of the challenges and a way around them.)
 
 Bugs resulting from invalid data can be tricky to detect and it can be time-consuming to clean up—best to prevent the possibility of it in the first place.
 
--- a/_styleguide/database.md
+++ b/_styleguide/database.md
@@ -14,7 +14,7 @@ PostgreSQL gives us a range of tools to ensure invalid data isn’t just unlikel
 
 - Use `UNIQUE` `CONSTRAINT`s/`INDEX`es (`unique: true` in migrations) to prohibit “duplicate” data. For instance, if it doesn’t make sense for a Lesson to be tagged with “Sales” _twice_, we want a unique composite index on the `tag` and `taggable`.
 - Use `NOT NULL` (`null: false` in migrations) to guarantee necessary relationships (e.g. make `company_id` non-nullable on tables that must be associated with a company) or attributes (if a `title` is required, make it `NOT NULL`).
-- **Update:** we no longer recommend adding `FOREIGN KEY` constraints by default, as we’ve run into performance issues on tables like `users` that have very many of them. However, we should still consider them in cases where referential integrity is essential. (And be mindful of performance when adding foreign key constraints: see [this migration](https://github.com/lessonly/lessonly/pull/5390/files#diff-bf0c0f95fbacafc133567b5e36e2f289R20) for an illustration of the challenges and a way around them.)
+- **Update:** we no longer recommend adding `FOREIGN KEY` constraints, as we’ve run into performance issues on tables like `users` that have very many of them. So avoid adding foreign-key constraints unless there is an overriding concern about data fidelity to justify them. (If so, be mindful of performance when adding foreign key constraints: see [this migration](https://github.com/lessonly/lessonly/pull/5390/files#diff-bf0c0f95fbacafc133567b5e36e2f289R20) for an illustration of the challenges and a way around them.)
 
 Bugs resulting from invalid data can be tricky to detect and it can be time-consuming to clean up—best to prevent the possibility of it in the first place.
 


_styleguide/database.md
```diff
@@ -8,13 +8,13 @@ main: true
...
-- Use `FOREIGN KEY` constraints to ensure that references to other tables like `company_id` point to records that actually exist, and specify `ON DELETE` `CASCADE` or `NULLIFY` behaviors when those records are removed. (Be mindful of performance when adding foreign key constraints: see [this migration](https://github.com/lessonly/lessonly/pull/5390/files#diff-bf0c0f95fbacafc133567b5e36e2f289R20) for an illustration the challenges and a way around them.)
+- **Update:** we no longer recommend adding `FOREIGN KEY` constraints by default, as we’ve run into performance issues on tables like `users` that have very many of them. However, we should still consider them in cases where referential integrity is essential. (And be mindful of performance when adding foreign key constraints: see [this migration](https://github.com/lessonly/lessonly/pull/5390/files#diff-bf0c0f95fbacafc133567b5e36e2f289R20) for an illustration of the challenges and a way around them.)
```
haleymnatz, line 17: Do you think we could add a little clarity on what "essential" means here?

stevegrossi: I'd like to clarify this, but honestly I can't really think of any situations where referential integrity would be "essential". I wanted to leave room for using them if we have a reason to, but maybe we should leave this out. I'll see if I can think of one 🤔

rreinhardt9: I'm kind of puzzling on this too 🤔 I was reading [this Thoughtbot example scenario](https://thoughtbot.com/blog/referential-integrity-with-foreign-keys) while I was thinking about it.

While I can't think of times where it's "essential", it seems like the advantage is more that you can be confident that if you have a record that reference another table by ID that the record will in fact be there when you go to retrieve it. It seems like it generally helps guard against something happening thin `delete_all` that wouldn't trigger Rails cascading deletes and could leave hanging records.

I wonder if it makes any difference when we "soft delete" records as opposed to direct deletion. Maybe if records in the table that is being referred to is never actually "destroyed" but rather just soft deleted in the application, it makes a foreign key constraint less important?

rreinhardt9:
@stevegrossi one question I have; do we know for sure that the foreign key constraint was the performance bottleneck or could it have been the "destroy" callbacks in Rails?

What I'm wondering is if we specify `on_delete: :cascade` on foreign key constraints but don't use the delete callbacks in the rails models, does that help us with performance while still allowing us to use foreign key constraints?

_styleguide/database.md
stevegrossi: > Maybe if records in the table that is being referred to is never actually "destroyed" but rather just soft deleted in the application, it makes a foreign key constraint less important?

I think so, which suggests we aren't getting much value out of our foreign-key constraints since data is so rarely truly deleted—and when it is, as in `DeleteCompany`, we're manually removing in batches to go easy on the DB rather than relying on `ON DELETE CASCADE` which deletes everything all at once.

stevegrossi:
> one question I have; do we know for sure that the foreign key constraint was the performance bottleneck or could it have been the "destroy" callbacks in Rails?

I believe when @sag and @wthenrichs noticed that deleting users was taking 20+ seconds each, it was from raw SQL, so definitely at the DB level rather than slowness in Rubyland.

> What I'm wondering is if we specify on_delete: :cascade on foreign key constraints but don't use the delete callbacks in the rails models, does that help us with performance while still allowing us to use foreign key constraints?

The problem we had with the 60+ constraints on `users.id` is that any time we wanted to hard-delete a user (e.g. in `DeleteCompany`), the database had to check, e.g. "are there any rows where `lessons.creator_id` is this user? No? Ok, move on.", "are there any progresses where `progresses.user_id` is this user? No? Ok, move on." and so on for every table that has a `FOREIGN KEY` constraint pointing to `users.id`. So regardless of what we have set up in our Rails models, having any kind of constraint in the DB on `users.id` (and so many of them) made it impossible to delete users at the scale we need.

rreinhardt9 (dismissed):
Gotcha!!! Thanks for the info Steve! That makes sense; I can't think of any good cases to be made for NOT adding foreign key constraints by default going forward based on the info you shared above :tada: 

This seems like a reversible decision; meaning that we can add the foreign key constraints later if not adding them causes unforeseen issues. I think that reversibility makes it a safe thing for us to try given we have some real pain coming from using them currently!

_styleguide/database.md
stevegrossi: Looking through all of our `FOREIGN KEY` constraints on `users`, I maybe see a distinction between which are "essential" and which are not, and it comes down to whether the data pointing to a user still makes sense if that user doesn't exist. For example, a `membership` record which relates a user to a group is meaningless if that user doesn't exist, so I might consider that an essential case for a FK constraint. On the other hand, we have columns like `lessons.creator_id`, `paths.last_updater_id`, `library_contents.added_by_id` where the relationship to a user is more like metadata: lessons, paths, and library contents are still functional if that user is gone. While there would still be benefits to constraints on those non-essential columns, if we need to ditch some FK constraints those are the ones I'd lose first.

rreinhardt9: I love this as a working definition Steve!!! Foreign keys for when record doesn't make sense apart from the associated record.

stevegrossi: @wthenrichs, @sag, @gowabash, unless we're 100% against foreign-key constraints, what would you think of [a guideline](#discussion_r534431937) that we consider them only in cases where the data is meaningless without the association? i.e. put a constraint on `memberships.user_id` but not `lessons.creator_id`? Would having significantly fewer FK constraints on the `users` table be good enough, or is our ultimate goal 0 of them referencing large tables?

gowabash: I would lean pretty hard towards don't add the FK ever.  Especially in the multi-tenant database world. You're not gaurenteeing correctness. And IMO, it should be `lessons.user_id` and you reference it as creator in the model, but that ship has sailed. 

I think the source of truth for data relationships for the overwhelming majority of Lessonly engineers is the rails model, not the database.  So the FK is genuinely not necessary.  

As data size increases, you need to think of the table as an independent item and the relationships are enforced in the application, not the database.

stevegrossi: > I would lean pretty hard towards don't add the FK ever.

I can agree with that, at least at the point we're at now where tables are getting larger and the cost of FKCs are increasing. That said, for the sake of discussion:

> Especially in the multi-tenant database world. You're not guaranteeing correctness.

Could you elaborate? I'm not clear on how multi-tenant impairs the referential integrity guarantees of FK constraints. (Or am I misunderstanding what you mean by correctness?)

> And IMO, it should be lessons.user_id and you reference it as creator in the model, but that ship has sailed.

We have both `lessons.creator_id` and `lessons.last_updater_id`...they both reference `users` but can't both be `lessons.user_id`, right? (Though if I hop on my own hobby horse about splitting up tables I could agree that these should be `lesson_creators.user_id` and `lesson_updaters.user_id` 😉)

> I think the source of truth for data relationships for the overwhelming majority of Lessonly engineers is the rails model, not the database. So the FK is genuinely not necessary.

I'd be more comfortable with that if we didn't so often bypass model validations (with `Model.import(data, validate: false)` for the sake of performance. I see model validations as a way to provide human-readable errors in the UI, not as bona fide guarantees of data fidelity. Obviously, constrain all the FKs isn't sustainable either given our current architecture, which has me thinking (longer term, not immediately) about things like:

- **what could make FKCs more performant?** Less data in tables. Maybe sharding large customers could eventually help here?
- **can we avoid skipping validations so often when bulk-importing lots of data?** Creating `Progress`es upon assignment is a barely-necessary hack we've been building upon for 5 years, if we stop doing that we remove the need to insert possibly hundreds of thousands of progresses at once and can maybe avoid the need for `import(validate: false)`.

gowabash: > Especially in the multi-tenant database world. You're not guaranteeing correctness.

You only make sure that it is a user you are connecting to.  They could be in the wrong company.  The multi-tenant aware key is really (id & company_id)

and in regards to validators, that's an opportunity for refactor.  There is more than 1 way to validate your input.  If you have to disable the feature because of volume, the feature needs looked at.

stevegrossi: > You only make sure that it is a user you are connecting to. They could be in the wrong company. The multi-tenant aware key is really (id & company_id)

Thanks for clarifying: I understand and agree. I dream of full segmentation with something [like postgres schemas](https://www.postgresql.org/docs/9.3/ddl-schemas.html) but I've yet to see a scalable solution with those.

> If you have to disable the feature because of volume, the feature needs looked at.

👍

stevegrossi: > I would lean pretty hard towards don't add the FK ever

I'll update the language to recommend never adding foreign-key constraints unless we have a specific overriding concern to justify them, since the concrete benefits are rare and the costs are constant, if variable.

stevegrossi:
For those who've been following the conversation, on Monday I announced updating our Style Guide recommendation on foreign-key constraints to:

> avoid adding foreign-key constraints unless there is an overriding concern about data fidelity to justify them.

I'd like to wrap this discussion up this week (meaning today) and would be grateful for a merge if you agree with the change.

gowabash (approved):

stevegrossi:
Thanks for the approval @gowabash 👍 Style Guide PRs don't go through QA or anything so I'll go ahead and merge this.

PR #142, opened 2020-12-04
Add Section about Feature Flags to Style Guide

b-ee:
This resolves [ch60063](https://app.clubhouse.io/lessonly/story/60063/add-section-about-feature-flags-to-style-guide), and gives us a way to organize Style Guide information related to Feature Flags!

### DIFF
--- a/_styleguide/feature_flags.md
+++ b/_styleguide/feature_flags.md
@@ -37,27 +37,40 @@ company.disable_feature!(:feature_name)
 # New Way, Recommended
 # This way is favored because we can perform feature checks on more types of objects.
 # The "old way" only allowed for company, now we can also pass in users & groups.
-AccessControl::Features::ExamplePolicy.new(company).feature_name?
+AccessControl::Features::ExamplePolicy.new(user).feature_name?
 AccessControl::Features::ExamplePolicy.new(company).grant!(:feature_name)
 AccessControl::Features::ExamplePolicy.new(company).revoke!(:feature_name)
 ```
 
-#### When writing unit tests, prefer stubbing feature flag checks
+#### Favor Checking Features on the Most-Specific Level
+
+```ruby
+# Not recommended, very broad
+AccessControl::Features::ExamplePolicy.new(company).feature_name?
+
+# Recommended!
+# This goes for Policies & Permissions as a whole.
+AccessControl::Features::ExamplePolicy.new(user).feature_name?
+```
+
+#### When Writing Unit Tests, Prefer Stubbing Feature Flag Checks
 
 This does not apply to integration tests!
 [See Testing Style Guide](https://about.lessonly.engineering/styleguide/testing/#avoid-stubbing-and-mocking-in-integration-tests) for more information on why.
 
 ```ruby
-# Not Ideal
-# Actually enabling the feature flag might cause issues in other tests!
+# Fine, But Not Ideal
 before do
   AccessControl::Features::ExamplePolicy.new(company).grant!(:feature_name)
 end
 
 # Recommended
+# This can improve performance and reduce potential issues.
 before do
-  allow(AccessControl::Features::ExamplePolicy).to receive(:new).with(company)
-    .and_return(double(feature_flag_name?: true))
+  example_policy_double = instance_double(AccessControl::Features::ExamplePolicy)
+  expect(example_policy_double).to receive(:feature_flag_name).and_return(true)
+
+  expect(AccessControl::Features::ExamplePolicy).to receive(:new).with(company).and_return(example_policy_double)
 end
 ```
 
--- a/_styleguide/feature_flags.md
+++ b/_styleguide/feature_flags.md
@@ -42,7 +42,10 @@ AccessControl::Features::ExamplePolicy.new(company).grant!(:feature_name)
 AccessControl::Features::ExamplePolicy.new(company).revoke!(:feature_name)
 ```
 
-#### When writing tests, prefer stubbing feature flag checks
+#### When writing unit tests, prefer stubbing feature flag checks
+
+This does not apply to integration tests!
+[See Testing Style Guide](https://about.lessonly.engineering/styleguide/testing/#avoid-stubbing-and-mocking-in-integration-tests) for more information on why.
 
 ```ruby
 # Not Ideal
--- a/_styleguide/feature_flags.md
+++ b/_styleguide/feature_flags.md
@@ -6,6 +6,14 @@ main: true
 
 ### Feature Flags
 
+#### Create Uniquely-Named Feature Flags
+
+Even when features live in different policies, they must all have unique names.
+
+#### Add Feature Policy Mapping to `AccessControl::Features::FeaturePolicies`
+
+The `FEATURE_POLICY_MAPPINGS` object maps feature names to the policies in which they can be found, which is needed in some areas of the application. Features are added to this mapping alphabetically.
+
 #### Add Translation Key for New Features
 
 ```ruby
@@ -33,3 +41,9 @@ AccessControl::Features::ExamplePolicy.new(company).feature_name?
 AccessControl::Features::ExamplePolicy.new(company).grant!(:feature_name)
 AccessControl::Features::ExamplePolicy.new(company).revoke!(:feature_name)
 ```
+
+#### Front End Implementation
+
+There is documentation in the React Style Guide which outlines how the `FeatureFlag` component can be used to do feature checks on React pages.
+
+[For more information, please visit that section of the Style Guide.](https://about.lessonly.engineering/styleguide/react/#featureflag-interface--in-transition---frontend-experiment)
--- a/_styleguide/feature_flags.md
+++ b/_styleguide/feature_flags.md
@@ -1,0 +1,35 @@
+---
+layout: styleguide
+title: Feature Flags Style Guide
+main: true
+---
+
+### Feature Flags
+
+#### Add Translation Key for New Features
+
+```ruby
+# Path to feature-related translation keys in config/locales/en.yml
+en:
+  admin:
+    company_details:
+      features:
+        feature_name: 'Classification: Name of Feature'
+# Without this addition, feature_name will be shown in Active Admin as "Feature Name".
+```
+
+#### Favor Calling Feature Checks & Toggles Directly on Policy
+
+```ruby
+# Old Way, Avoid
+company.has_feature?(:feature_name)
+company.enable_feature!(:feature_name)
+company.disable_feature!(:feature_name)
+
+# New Way, Recommended
+# This way is favored because we can perform feature checks on more types of objects.
+# The "old way" only allowed for company, now we can also pass in users & groups.
+AccessControl::Features::ExamplePolicy.new(company).feature_name?
+AccessControl::Features::ExamplePolicy.new(company).grant!(:feature_name)
+AccessControl::Features::ExamplePolicy.new(company).revoke!(:feature_name)
+```
--- a/_styleguide/feature_flags.md
+++ b/_styleguide/feature_flags.md
@@ -42,6 +42,22 @@ AccessControl::Features::ExamplePolicy.new(company).grant!(:feature_name)
 AccessControl::Features::ExamplePolicy.new(company).revoke!(:feature_name)
 ```
 
+#### When writing tests, prefer stubbing feature flag checks
+
+```ruby
+# Not Ideal
+# Actually enabling the feature flag might cause issues in other tests!
+before do
+  AccessControl::Features::ExamplePolicy.new(company).grant!(:feature_name)
+end
+
+# Recommended
+before do
+  allow(AccessControl::Features::ExamplePolicy).to receive(:new).with(company)
+    .and_return(double(feature_flag_name?: true))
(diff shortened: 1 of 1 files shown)


_styleguide/feature_flags.md
```diff
@@ -0,0 +1,35 @@
+---
+layout: styleguide
+title: Feature Flags Style Guide
+main: true
+---
+
+### Feature Flags
```
case-eee, line 7: In the past, I've seen folks forget to add the feature flag declaration in all the required places..

The first place is the `access_control/features` specific policy. For example, in `policies/access_control/features/in_person_training_policy.rb`, we'd need it listed here:

```ruby
module AccessControl
  module Features
    class InPersonTrainingPolicy < AccessControl::Features::Base
      actions(
        training_events: { action: 1, tags: [:pro, :default] },
        events_v2: { action: 2, tags: :deprecated },
        # ....
        events_attendee_limit: { action: 13, tags: :deprecated },
        events_zoom_and_other_video_conferencing_meetings: { action: 14, tags: :development }
      )
    end
  end
end
```

The second place is the `policies/access_control/features/feature_policies.rb` file in the long list of features.

Do you think we should also add notes about these here?

_styleguide/feature_flags.md
```diff
@@ -0,0 +1,35 @@
...
+# Path to feature-related translation keys in config/locales/en.yml
+en:
+  admin:
+    company_details:
+      features:
+        feature_name: 'Classification: Name of Feature'
+# Without this addition, feature_name will be shown in Active Admin as "Feature Name".
+```
+
+#### Favor Calling Feature Checks & Toggles Directly on Policy
```
case-eee, line 21: This is specific to the backend, I wonder if we want to also include a note on how to check feature flags on the frontend? @tdmiller1 may have some insight here 🎉

case-eee (changes requested):
@b-ee I love it! Thanks again for taking this on 🥳 Typically when I have additions to the Style Guide, I also throw it out in #dev-team just in case anyone else has thoughts/suggestions. What do you think about also getting some additional 👀 on this?

_styleguide/feature_flags.md
b-ee: There's actually a thing in the React section about that, Tucker showed me yesterday:
https://about.lessonly.engineering/styleguide/react/#featureflag-interface--in-transition---frontend-experiment

I could maybe just link to that section?

_styleguide/feature_flags.md
b-ee: I think we could note that, sure! Sometimes the policy mapping is left off, which makes things not work later.

_styleguide/feature_flags.md
case-eee: That works!

_styleguide/feature_flags.md
case-eee: I've seen this a few times so better safe than sorry 😄

_styleguide/feature_flags.md
```diff
@@ -0,0 +1,49 @@
...
```
gardengeek, line 29: @b-ee I wonder if it would be useful to include the recommended way to stub feature flags in the test (so that they don't actually enable/disable a ff & cause issues later).

b-ee: I considered that, but I've seen enough variations that I hadn't felt confident about being decisive on this item.  😅 

This is usually my go-to method, do you think this would be a good example snippet to put in the Style Guide?

```
allow(AccessControl::Features::ExamplePolicy).to receive(:new).with(company)
    .and_return(double(feature_flag_name?: true))
```

b-ee: https://github.com/lessonly/lessonly.github.io/pull/142/commits/c251464bd6f9a35b3ffea52dc968116f460ed75b

_styleguide/feature_flags.md
```diff
@@ -0,0 +1,65 @@
...
+
+### Feature Flags
+
+#### Create Uniquely-Named Feature Flags
+
+Even when features live in different policies, they must all have unique names.
+
+#### Add Feature Policy Mapping to `AccessControl::Features::FeaturePolicies`
+
+The `FEATURE_POLICY_MAPPINGS` object maps feature names to the policies in which they can be found, which is needed in some areas of the application. Features are added to this mapping alphabetically.
```
haleymnatz, line 15: One thing that could be nice to add here is the file path where this is located, just in case someone doesn't know where this is and doesn't think to do a global search for `FEATURE_POLICY_MAPPINGS`.

_styleguide/feature_flags.md
```diff
@@ -0,0 +1,65 @@
...
+company.has_feature?(:feature_name)
+company.enable_feature!(:feature_name)
+company.disable_feature!(:feature_name)
+
+# New Way, Recommended
+# This way is favored because we can perform feature checks on more types of objects.
+# The "old way" only allowed for company, now we can also pass in users & groups.
+AccessControl::Features::ExamplePolicy.new(company).feature_name?
+AccessControl::Features::ExamplePolicy.new(company).grant!(:feature_name)
+AccessControl::Features::ExamplePolicy.new(company).revoke!(:feature_name)
```
joshmfrankel, line 42: For the example could we switch out `company` for `user`? This is generally the preferred way of working with features/permissions as it allows for a future where we could have granular permissions on the user level but not company level :-)

This same thought throughtout the doc :-)

haleymnatz (commented):
This is 🔥

_styleguide/feature_flags.md
```diff
@@ -0,0 +1,65 @@
...
+```
+
```
joshmfrankel, line 44: Could we add a section regarding permissions on specific objects? For example, you can have blanket permissions that cover the entire product or granular record specific permissions:

```
AccessControl::RolePolicy.new(@role).grant!(:manage, creator_role)
``` 

The `.grant!(:manage, creator_role)` part is the one it'd be good to get an example of :-)

_styleguide/feature_flags.md
```diff
@@ -0,0 +1,65 @@
...
+#### When writing tests, prefer stubbing feature flag checks
+
+```ruby
+# Not Ideal
+# Actually enabling the feature flag might cause issues in other tests!
```
joshmfrankel, line 49: > Actually enabling the feature flag might cause issues in other tests!

Could you explain this part a bit more? I understood this format as being acceptable in some circumstances and wasn't aware that it would bleed context between tests

_styleguide/feature_flags.md
b-ee: It's in the header! Do you think `AccessControl::Features::FeaturePolicies` is descriptive enough?

_styleguide/feature_flags.md
b-ee: [Chris left a comment about this!](https://github.com/lessonly/lessonly.github.io/pull/142#discussion_r537684430)

I made the change based on her feedback, but I did also understand this as the standard myself. I can't personally think of situations where enabling the feature flag caused flakes.

As I noted in my response to Chris, I hadn't added this section in the first place because I wasn't super confident that there WAS a standard.. so maybe there isn't, if you think it's fine either way?

b-ee: Maybe if we decide this IS a good standard to favor, it would help to encourage consistency. We have 1000 different ways of handling feature flags in tests, sometimes it's a pain finding reference code.  🙃

_styleguide/feature_flags.md
b-ee: I agree that this could be helpful, but do you think it's starting to get a little off-topic? Deals less with feature flags, more with Policies & Permissions as a whole.

(I'm really asking, not pushing back. Policies/Permissions are a lot fuzzier for me than FFs, lol)

b-ee: If they ARE different enough that maybe they don't belong here... that's definitely not to say they shouldn't have a home somewhere. Maybe somewhere else in the Style Guide, or maybe this section gets rebranded a bit. I think that might be a little out of scope for now, though.

_styleguide/feature_flags.md
b-ee: Kind of relates to another reply I left for you ... I like using `company` here in the context of feature flags because they're generally enabled at the company level.

When we're checking a Feature policy, do we also pass in `user`? Or does it make just as much sense to pass in `company` in this context because it's a company-level feature flag?

_styleguide/feature_flags.md
rreinhardt9: I the the main reason that we've tended to stub the polices was just for performance; instead of actually having the test check the permission from the DB it just hits the stub. In that way, it's great to stub it when it makes sense for performance but there also may be times when granting it directly might make sense for what you are testing. Sometimes with things like feature specs I've found it helpful to just grant the features directly that are needed.

I would caution against using a generic double in your stub, I've done that in the past and it can be very hard to figure out what double isn't getting called with the expected message sometimes (you'll get a generic "stub receive unexpected message" warning). Additionally, a generic double doesn't check if the class it's stubbing actually implements the method being stubbed. Because of that, I've used `instance_double` for this purpose:
```
example_policy_double = instance_double(AccessControl::Features::ExamplePolicy)
expect(example_policy_double).to receive(:feature_flag_name).and_return(true)

allow(AccessControl::Features::ExamplePolicy).to receive(:new).with(company).and_return(example_policy_double)
```

_styleguide/feature_flags.md
```diff
@@ -0,0 +1,65 @@
...
+layout: styleguide
+title: Feature Flags Style Guide
+main: true
+---
+
+### Feature Flags
+
+#### Create Uniquely-Named Feature Flags
+
+Even when features live in different policies, they must all have unique names.
```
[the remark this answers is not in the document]
b-ee: _Mmhmmm._   🙄   It sucks to have the extra policy mapping step, and to have to care about the unique names.  _Somedayyyy..._   🤞

_styleguide/feature_flags.md
b-ee: Hmm okay. I've never had that issue (with the double not being called, but not having clarity around what's breaking), but good to know it's something that COULD happen.

Well. It sounds like there are two options:

1) Fine-tuning the suggested code
2) Letting people do whatever they feel is best, and removing this section of the Style Guide completely

_styleguide/feature_flags.md
case-eee: @joshmfrankel I'm curious - should we all be passing `user` into these checks instead of `company` right now? I know we aren't at the point where we're enabling features based on the individual use, but we might at some point in the future. Since feature flags can only be enabled at the company, do you think it'd be confusing for folks if we asked them to check whether or not a user has a feature flag enabled?

My hunch is no - but I'd love for us to be consistent moving forward with how we're checking feature flags!

b-ee: Also, the comment above these lines DOES mention that we _can_ use users and groups. So this functionality is still mentioned here, even if it's not quite normal practice yet!

rreinhardt9: I agree with the use of "user" being the standard we'd want to set for the first line of this example `AccessControl::Features::ExamplePolicy.new(user).feature_name?` 

The the hope behind introducing Accessly was that you could grant permission at multiple levels wherever you check the permission would not have to care about what level permission was granted at. So as a rule of thumb, you want to *check* for permission at the "lowest" level ("user" for us) even if you are *granting* permission typically at the "company" level. That keeps the door open for us to, say, grant a feature to a user directly or through a group. Practically speaking, you should probably just always be checking at the user level unless there is a specific reason to check at a "higher" level.

b-ee: I can get behind that! I'll change the JUST first line to check for `user`, that makes sense.

_styleguide/feature_flags.md
b-ee: I might just rebrand it as a suggestion for performance.  🤷‍♀️

_styleguide/feature_flags.md
b-ee: There's a conversation happening in this thread:   https://github.com/lessonly/lessonly.github.io/pull/142#discussion_r537720975

It seems like different people have different opinions, so if I keep this section, I probably won't do so as a **Standard**. Just as a suggestion!!

_styleguide/feature_flags.md
rreinhardt9: Yeah, that seems like a good idea! That way, it will still be there to help provide guidance on how any why they might want to stub a policy check in Lessonly so the engineer can make an informed decision.

_styleguide/feature_flags.md
case-eee: I like this!

case-eee (approved):
@b-ee Thanks for making taking the initiative here, getting the ball moving, and collaborating with folks on feedback 🥳 Let's get this merged!

_styleguide/feature_flags.md
b-ee: I also added a section to make this ultra-clear!

***Favor Checking Features on the Most Specific Level***

RickGrey (approved):

PR #143, opened 2021-01-28
Adds /search page which searches our wikis on Github

stevegrossi:
Screen Shot 2021-01-27 at 7 00 17 PM

## Why

We keep some of our internal technical documentation in Github wikis, and with [this initiative](https://app.clubhouse.io/lessonly/story/60251/where-should-documentation-live-for-technical-topics) we plan on moving much more of our current and future documentation over to Github Wikis so that it's easier to update, accessible online without having to clone a repo, and—most relevant here—globally searchable across all our repos. However, that requires three steps: visit github.com/lessonly, enter your search term in the nav bar search, and click "Wikis" in the results to narrow them down. I'd like to make that a little bit easier by providing a bookmarkable form that searches our wiki content directly.

## What

This change will add a search form (pictured above) at about.lessonly.engineering/search which submits its search query directly to Github pre-filtered to Lessonly repos' wiki content. This results page will automatically inherit the access privileges of the Github user, returning only public content (e.g. `accessly`, our style guide) for non-Lessonly-employees and any internal content a logged-in Lessonly developer has access to.

## Testing Notes

- See [the README](https://github.com/lessonly/lessonly.github.io#running-it-locally) for running this app locally
- `git checkout stevegrossi/llamapedia`
- visit http://localhost:4000/search or click "Search" in the header nav from any page
- View out the results on Github


case-eee (approved):
@stevegrossi This works as expected ✅ And looks fantastic. Let's get this merged!

PR #145, opened 2021-05-28
Update search button cursor

emmoistner:
# Description

Minor styling update the search button have `cursor: pointer` instead of the default cursor.


stevegrossi (approved):
Thanks for this! I'm thought the `<input />` element would use a pointer by default since it's interactable but I was mistaken.

PR #146, opened 2021-06-22
Clarify that we do not recommend using foreign-key constraints and what to use instead

stevegrossi:
While we have many `FOREIGN KEY` constraints in our database, we have been removing them from large, commonly-referenced tables like `users` because Postgres implements them as triggers which can slow down deleting records from the referenced table. And while `ON DELETE CASCADE` constraints can help with cleaning up data incrementally, we prefer to use Rails' `dependent:` relationships for incremental cleanup, and our `DeleteCompany` service for cleaning up an entire account.

There was still some confusion with the earlier guidance, so I want to update it to make it clear that we recommend never adding `FOREIGN KEY` constraints as well as offer the preferred alternative (Rails `dependent:` callbacks on associations).

### DIFF
--- a/_styleguide/database.md
+++ b/_styleguide/database.md
@@ -10,13 +10,11 @@ The PostgreSQL Wiki has a handy [“Don’t Do This”](https://wiki.postgresql.
 
 ## Enforce data integrity with unique indexes and not-null constraints
 
-PostgreSQL gives us a range of tools to ensure invalid data isn’t just unlikely but impossible:
+Bugs resulting from invalid data can be tricky to detect and it can be time-consuming to clean up—best to prevent the possibility of it in the first place. PostgreSQL gives us a range of tools to ensure invalid data isn’t just unlikely but impossible:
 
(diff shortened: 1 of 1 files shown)


_styleguide/database.md
```diff
@@ -10,13 +10,11 @@ The PostgreSQL Wiki has a handy [“Don’t Do This”](https://wiki.postgresql.
 
 ## Enforce data integrity with unique indexes and not-null constraints
 
-PostgreSQL gives us a range of tools to ensure invalid data isn’t just unlikely but impossible:
+Bugs resulting from invalid data can be tricky to detect and it can be time-consuming to clean up—best to prevent the possibility of it in the first place. PostgreSQL gives us a range of tools to ensure invalid data isn’t just unlikely but impossible:
```
stevegrossi, line 13: I moved this up because it helps to know why it matters _before_ you start reading about something.

mecrowell (approved):

PR #149, opened 2021-09-30
Strengthens our recommendation against string interpolation in raw SQL

stevegrossi:
Screen Shot 2021-09-30 at 10 07 51 AM

While we formerly recommended preferring ActiveRecord code over raw SQL, we now require avoiding raw SQL entirely when string interpolation is involved. Because even if we take steps to do it safely today (only ever passing known-safe values), that doesn't guarantee it will always be used safely in the future. It is better to make our code secure by default.


nathan-salazar (approved):
Looks good to me!

